PR #203, opened 2024-01-03
Radiology report exports

### LINKED ISSUE #161: Export radiology reports as parquet file

The EHR-API pipeline's de-identified radiology reports will be written as a parquet file to a directory on the GAE mounted into the container.

The parquet file should have the following columns:
* De-IDed report text
* procedure_occurence_id (aka. EHR imaging identifier)
* DICOM link (ie. the hashing API’s response for {patient mrn}{accession number})

### Testing

Integration test of input imaging messages -> parquet written in the expected location. Imagine having a test directory which is controlled by a fixture: setup - create dir, teardown - delete directory. 

### Dependencies

Export directory definition from:

- [x] https://github.com/UCLH-Foundry/PIXL/issues/162

#### Current state

Patient data is queried from EMAP and added to a postgres table with the de-identified radiology reports

The radiology report is found in `report_text` in `PatientEHRData`.

#### Details

OMOP ES would have already created an OMOP export which will be copied to the [export directory](https://github.com/UCLH-Foundry/PIXL/pull/182), the imaging reports should be written here. 

We're not addressing this in the current ticket but for context: After this step, there could be a manual check that the de-identification is performing as expected (e.g. check for data drift), then the CLI can be used to POST the images to the DSH.

jeremyestein:
Define an API call in pixl ehr that triggers:
 * extract all EHR records from the PIXL DB that match the specified extract (as defined by project name and extract datetime)
 * dump to a single parquet file in the mounted export dir

For a later PR:
* Adding a command to the CLI to call this API
* Add to system test (depends on CLI call) - (add cli call after pixl populate, but how to wait until done processing)

Fixes #161


pixl_ehr/src/pixl_ehr/main.py
```diff
@@ -58,6 +63,21 @@ async def startup_event() -> None:
...
+@app.post(
+    "/export-radiology-as-parquet",
+    summary="Copy all radiology reports in the PIXL DB to a parquet file",
+)
+def export_radiology_as_parquet(project_name: str, extract_datetime: datetime) -> None:
+    """Batch export of all radiology reports in PIXL DB to a parquet file."""
+    # This command has to tell us the extract identifiers so we can extract to the correct
+    # path, however we can't verify that the reports in the DB actually came from that extract.
+    # Also we can't check that all reports in the queue have been processed.
+    # So we are relying on the user passing correct parameters here.
```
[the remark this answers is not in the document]
stefpiatek: I _think_ that's what we landed on doing so the comment may be out of date, or I've misunderstood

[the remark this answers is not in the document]
jeremyestein: I don't know what we agreed in the end. Should they not go in the `extract` table since they are properties of an extract? But then we don't link the `ehr_*` tables to `image` or `extract` (perhaps because the images might not exist yet - but the extracts will!) so it could be hard.
Relatedly, can we move the contents of `pixl-db_init.sh` to `create_pixl_tbls.py` - would be nice to have the DB schema all in one place.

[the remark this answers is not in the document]
stefpiatek: > I don't know what we agreed in the end. Should they not go in the extract table since they are properties of an extract? But then we don't link the ehr_* tables to image or extract (perhaps because the images might not exist yet - but the extracts will!) so it could be hard.

I think for now, we should pretend like the `extract` table doesn't exist so that we can control it independently, a later stage would be rationalisting (and normalising) it in sqlaclhemy later on. We should make sure we store each `ehr_` row with the `project_name`  and `extract_datetime`.

>  Relatedly, can we move the contents of pixl-db_init.sh to create_pixl_tbls.py - would be nice to have the DB schema all in one place.

Yes though, I'd say that this is a later task if we get everything else that we need done and have extra time.

pixl_ehr/src/pixl_ehr/main.py
```diff
@@ -58,6 +62,21 @@ async def startup_event() -> None:
...
+)
+def export_radiology_as_parquet(project_name: str, extract_datetime: datetime) -> Path:
+    """Batch export of all radiology reports in PIXL DB to a parquet file."""
+    # This command has to tell us the extract identifiers so we can extract to the correct
+    # path, however we can't verify that the reports in the DB actually came from that extract.
+    # Also we can't check that all reports in the queue have been processed.
+    # So we are relying on the user passing correct parameters here.
+    anon_data = PIXLDatabase().get_radiology_reports()
+    pe = ParquetExport(project_name, extract_datetime)
+    return pe.export_radiology(anon_data)
```
jeremyestein, line 77:
```
error: Returning Any from function declared to return "Path"  [no-any-return]
```
Why this error when `pe.export_radiology` *is* annotated to return `Path`?

stefpiatek: Yeah I'm not sure about that one, if you look for `export_radiology` in the `.mypy_cache` then it does have the return type as pathlib.Path

```
"export_radiology": {
            ".class": "SymbolTableNode",
            "kind": "Mdef",
            "node": {
              ".class": "FuncDef",
              "abstract_status": 0,
              "arg_kinds": [
                0,
                0
              ],
              "arg_names": [
                "self",
                "anon_data"
              ],
              "dataclass_transform_spec": null,
              "flags": [],
              "fullname": "pixl_core.src.core.omop.ParquetExport.export_radiology",
              "name": "export_radiology",
              ...
                "ret_type": "pathlib.Path",
                ...
```

Might just need to manually set it because clearly something in mypy isn't quite right

```suggestion
    parquet_file: Path = pe.export_radiology(anon_data)
    return parquet_file
```

stefpiatek: 👌 the API wraps around a method which returns the value but the API method doesn't?

hasher/src/hasher/endpoints.py
```diff
@@ -57,3 +57,12 @@ async def hash_accession_number(message: str) -> Response:
...
     return Response(content=generate_hash(message, 64), media_type="application/text")
+
+
+@router.get(
+    "/hash-image-identifier",
+    summary="Produce secure hash appropriate for an image identifier "
+    "(MRN + accession num concatenated)",
+)
+async def hash_image_identifier(message: str) -> Response:
+    return Response(content=generate_hash(message, 64), media_type="application/text")
```
stefpiatek, line 68: Oh this surprises me. Why are we making a seperate endpoint that has the same implementation as `hash`?

pixl_core/src/core/omop.py
```diff
@@ -17,24 +17,38 @@
...
 
-    def __init__(self, root_dir: pathlib.Path = root_from_install) -> None:
-        """Create instance of OMOPExtract helper."""
-        self.export_dir = root_dir / "exports"
+    def __init__(self, project_name: str, extract_datetime: datetime.datetime) -> None:
+        """
+        :param project_name: name of the project
+        :param extract_datetime: datetime that the OMOP ES extract was run
+        :param root_dir: export directory root
+        """
```
stefpiatek, line 44: I like this change a lot. Much more clean to do set this up as part of the intialisation

pixl_core/src/core/omop.py
```diff
@@ -65,24 +71,52 @@ def copy_to_exports(
...
-        return project_slug
+        latest_public.symlink_to(self.public_output, target_is_directory=True)
+        return self.project_slug
+
+    def export_radiology(self, anon_data: list[tuple[Any, ...]]) -> pathlib.Path:
+        """Export radiology reports to parquet file"""
+        # The parquet file should have the following columns:
+        # - De-IDed report text
+        # - procedure_occurrence_id (aka. EHR imaging identifier)
+        # - DICOM link (ie. the hashing API's response for {patient mrn}{accession number})
```
stefpiatek, line 95: These can probably be removed

pixl_core/src/core/omop.py
```diff
@@ -65,24 +71,52 @@ def copy_to_exports(
...
+        parquet_file = self.radiology_output / "radiology.parquet"
+
+        parquet_header_names = ["image_identifier", "procedure_occurrence_id", "image_report"]
+        if anon_data and len(parquet_header_names) != len(anon_data[0]):
+            err_str = (
+                f"Passed in data must have expected number of columns "
+                f"(got {len(anon_data[0])}, expected {len(parquet_header_names)})"
+            )
+            raise ValueError(err_str)
+        export_df = pd.DataFrame(anon_data, columns=parquet_header_names)
```
stefpiatek, line 107: Is there a reason you went for making the dataframe here, and not being passed a dataframe that's been built directly from the query in the `ehr-api`? 

Feels like if the query changed (e.g. changing columns) we'd have to change it here, rather than changing it all in one place

pixl_ehr/pyproject.toml
```diff
@@ -17,6 +17,7 @@ dependencies = [
     "psycopg2==2.9.5",
     "azure-identity==1.12.0",
     "azure-storage-blob==12.14.1",
+    "pyarrow==14.0.1",
```
stefpiatek, line 20: is pyarrow needed here? I thought it would have been in pixl core which we're installing

pixl_ehr/src/pixl_ehr/main.py
```diff
@@ -58,6 +62,19 @@ async def startup_event() -> None:
...
 
 
+@app.post(
+    "/export-radiology-as-parquet",
+    summary="Copy all matching radiology reports in the PIXL DB to a parquet file",
+)
+def export_radiology_as_parquet(project_name: str, extract_datetime: datetime) -> None:
+    """Batch export of all matching radiology reports in PIXL DB to a parquet file."""
+    # NOTE: we can't check that all reports in the queue have been processed.
+    # So we are relying on the user waiting until processing has finished before running this.
```
stefpiatek, line 72: Might be worth having this in the docstring rather than in-code comment?

pixl_ehr/tests/test_processing.py
```diff
@@ -186,11 +257,81 @@ async def test_message_processing() -> None:
...
-    anon_mrn, anon_accession_number = anon_row[:2]
+    assert len(anon_all_rows) == 1
+    anon_row = anon_all_rows[0]
+    anon_mrn, anon_accession_number, anon_report_text = anon_row
     assert anon_mrn != mrn
-    assert anon_accession_number != accession_number
+    assert anon_accession_number != message.accession_number
+
+    # No deidentification performed so we expect the report to stay identical
+    assert report_text == anon_report_text
```
stefpiatek, line 270: yeah maybe we should make the mock cogstack add on something like `**DE-IDENTIFIED**` to the end of the string so that we know it has been processed in tests

pixl_ehr/tests/test_processing.py
```diff
@@ -186,11 +257,81 @@ async def test_message_processing() -> None:
...
+    latest_parquet_file.is_symlink()
+
+
+@pytest.mark.processing()
+@pytest.mark.asyncio()
+async def test_radiology_export_multiple_projects(example_messages) -> None:
+    """
+    GIVEN 4 messages, each from a different project+extract processed by the EHR API
+    WHEN export_radiology_as_parquet is called for 1 given project+extract
+    THEN only the radiology reports for that project are exported
```
stefpiatek, line 308: nice! Might be worth saying that there are two exports for the project at different datetimes so we know its checking that as well

postgres/pixl-db_init.sh
```diff
@@ -16,7 +16,7 @@
 set -euxo pipefail
 
 # Create the EHR schema and associated tables
-columns_and_types="mrn text, accession_number text, age integer, sex text, ethnicity text, height real, weight real, gcs integer, xray_report text"
+columns_and_types="mrn text, accession_number text, image_identifier text, procedure_occurrence_id integer, age integer, sex text, ethnicity text, height real, weight real, gcs integer, xray_report text, project_name text, extract_datetime timestamp"
```
stefpiatek, line 19: Is it worth tracking with timezone?

stefpiatek (commented):
Looking good, only blocking change is adding in a new endpoint to the hashing API, worth discussing that and then some other thoughts

hasher/src/hasher/endpoints.py
jeremyestein: Mainly I was following the existing pattern. But do the different types of input even need different hashing methods? I'm not sure what the requirements are here.

pixl_core/src/core/omop.py
jeremyestein: Although I've just noticed that `root_dir` no longer exists - our linting should be configured to catch this.

pixl_ehr/pyproject.toml
jeremyestein: Apparently yes! See #213 
```
E           ImportError: Unable to find a usable engine; tried using: 'pyarrow', 'fastparquet'.
E           A suitable version of pyarrow or fastparquet is required for parquet support.
E           Trying to import the above resulted in these errors:
E            - Missing optional dependency 'pyarrow'. pyarrow is required for parquet support. Use pip or conda to install pyarrow.
E            - Missing optional dependency 'fastparquet'. fastparquet is required for parquet support. Use pip or conda to install fastparquet.
```

postgres/pixl-db_init.sh
jeremyestein: Do we have that information? If so, then yes.

pixl_core/src/core/omop.py
jeremyestein: Yes, I think it's better to do it all together like that, so I'm moving it. Also removed the validation as it turns out pandas does that anyway. Only slight thing is that we now have some parquet extract-specific logic (the column names) in the `_databases.py` module, but eh.

pixl_ehr/tests/test_processing.py
jeremyestein: I agree but can we do it in a separate PR? There are quite a few other changes I want to make to testing that could be done together as a separate piece of work.
I would also suggest that we monkeypatch `deidentify_text` during unit tests so we don't need the fake cogstack service except for in integration/system tests, but then it might not be the only thing that needs monkeypatching...

postgres/pixl-db_init.sh
jeremyestein: From the example logfile:
```
"datetime":"2023-12-07 14:08:58",
```
This doesn't specify a timezone or an offset, although if we're sure it's UTC we could add that information ourselves.
However, do we use it as anything other than a fairly opaque identifier for an extract? If it's not a product output itself and we're not doing any actual calculations on it, I don't think we need to care about the timezone.

pixl_ehr/tests/test_processing.py
milanmlft: Oops, sorry @jeremyestein I already pushed this into this PR before seeing your comment

jeremyestein: No problem

pixl_core/src/core/omop.py
milanmlft, line 1: Probably better to rename this file as well? Something like `exports.py`?

hasher/src/hasher/endpoints.py
stefpiatek: Yeah I think the previous is a bit silly too. I'd remove and just use `/hash` I assume it was planned to do something like restricting the size of it like they did with accession number, but then that changed and they didn't remove the superflous extra endpoint

postgres/pixl-db_init.sh
stefpiatek: yeah that's fair, in that case happy with keeping it as-is, as we don't know or particularly care about the timezone other than an identifier

hasher/src/hasher/endpoints.py
jeremyestein: Have removed the new endpoint. Shall we save the consolidation of the other endpoints for another PR?

pixl_ehr/src/pixl_ehr/_processing.py
```diff
@@ -71,6 +74,10 @@ class PatientEHRData:
 
     mrn: str
     accession_number: str
+    image_identifier: str
+    procedure_occurrence_id: int
+    project_name: Optional[str]
+    extract_datetime: Optional[datetime]
```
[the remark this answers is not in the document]
milanmlft: I think so it wouldn't break any older code, if there would still be instances of `PatientEHRData` around that don't (need to) set those attributes. But I don't think that's the case, so I'll change it to be non-optional :)

hasher/src/hasher/endpoints.py
milanmlft: Maybe make a new issue so we don't forget / to discuss further?

PR #207, opened 2024-01-04
Filter procedures to ngtube and chest x-ray

stefpiatek:
Hardcoded for now, when we start procesing other data types we'll work out a more configurable way to define the expected concept ids. 

Updated to use @Johnnywatts's more realistic data that we've used in the integration test


cli/src/pixl_cli/_io.py
```diff
@@ -126,8 +126,10 @@ def _check_and_parse_parquet(private_dir: Path, public_dir: Path) -> pd.DataFram
     accessions = pd.read_parquet(private_dir / "PROCEDURE_OCCURRENCE_LINKS.parquet")
     # study_date is in procedure.procedure_date
     procedure = pd.read_parquet(public_dir / "PROCEDURE_OCCURRENCE.parquet")
+    # harcoded to ng tube and chest x-rays for now, will allow this to be configured in the future
+    imaging_procedures = procedure.loc[procedure.procedure_concept_id.isin([4163872, 42538241])]
```
milanmlft, line 130: Maybe add a `# TODO` and link to a new issue for this?

milanmlft (approved):
🚀

PR #214, opened 2024-01-11
Add tests for pseudonymisation 

stefpiatek:
# Addressed

- pulled in main, quite a few changes here but if you pull in main yourself then I think those will go away?
- Broke out some of the helper methods into other files so its a bit easier to have concise test files
- Using environmental variables instead of config file! Now I understand why it was complaining about not having the yaml file, all of the services will be passed environmental variables which are read in by `config` so that makes it easier and tests can be run from anywhere
- Had a quick go at writing some tests for the pipeline, I think they're sufficient for our changes

# Next steps assuming you're happy with these

I think this will need a couple of things for it to work:

- Add required enviornmental variables using something like `os.environ`, `conftest.py` is a classic place to do this because its run before the test. Then the env variables exist during testing with dummy values so we don't throw errors
- monkeypatch the requests.get method to return an object which has the `statuscode` and `content` fields
  - status code can always be 200
  - for a fake hash the tests are expecting adding dashes between each character and returning it encoded into bytes using utf-8


ruaridhg (approved):
Looks good, will add in the env vars now

PR #216, opened 2024-01-11
Fix ruff rule PLW2901 "redefine loop name"

### LINKED ISSUE #157: [Project PIXL]: review `ruff` rule `PLW2901` "redefine loop name"

Pass `ruff` rule PLW2901 by either fixing or ignoring.

### Testing

If we decide to rewrite the `for` loop, need to test whether the results are still valid.

```
ruff check --select PLW2901 .
```

### Documentation

_No response_

### Dependencies

_No response_

### Details and Comments

In https://github.com/UCLH-Foundry/PIXL/blob/46f5b3cfa47134a30516b36696c1d74a853a84ba/pixl_ehr/src/pixl_ehr/_queries.py#L36-L45 the `line` loop variable gets redefined during the for loop, which, according to `ruff`, [is bad](https://docs.astral.sh/ruff/rules/redefined-loop-name/). 

However, this may be intentional, in which case it would be justified to ignore the rule for this code block with `# noqa: PLW2901`. Needs further investigation.

milanmlft:
Just cleaning up some old ruff issues.
Closes #157

### DIFF
--- a/pixl_ehr/src/pixl_ehr/_queries.py
+++ b/pixl_ehr/src/pixl_ehr/_queries.py
@@ -41,17 +41,17 @@ class SQLQuery:
                 continue
 
             for key, value in context.items():
-                line = line.replace("${{ " + str(key) + " }}", str(value))  # noqa: PLW2901
+                new_line = line.replace("${{ " + str(key) + " }}", str(value))
 
-                n = line.count(f":{key}")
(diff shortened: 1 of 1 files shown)


milanmlft:
Note: auto-merge is on!

stefpiatek (approved):
🎉

PR #217, opened 2024-01-11
Fix error in placeholder substitution

jeremyestein:
I think this unbreaks things but I may have misunderstood what this code is supposed to do.


stefpiatek (approved):
Thank for fixing!

PR #226, opened 2024-01-16
Upload DICOM images with FTPS

### LINKED ISSUE #190: Export DICOM images via FTPS

De-identified DICOM images will be downloaded from orthanc-anon and sent via FTPS to the DSH under the project slug directory using the pseudonymised id as the filename (e.g. `{project-slug}/{study-pseduonymised-id}.zip`)

Breakdown of steps for progress:

- [x] get pseudonymised identifier for image from DICOM data (currently in patient ID tag  you can do this in a REST request: `{dicom-server}/instances/{resource-id}/content/0010-0020` gets us just the patient idenfier, see below for details)  
- [x] get project name slug from PIXL db using the pseudonymised identifier (in pixl db: `Image.hashed_identifier`)
- [x] download zip of DICOM data (see details below)
- [x] upload zip of DICOM data to `{project-slug}/{study-pseduonymised-id}.zip` on the DSH (see details and prototype repo below)
- [x] Update the existing image row in the database with the current time added to `exported_at`

### Testing

In an ideal world we'd have a mock DSH in docker compose which we can ftp into. If we can set this up in a day then probably worth it. Have a prototype repo for setting this up so I _think_ this should be quick,

### Current State

At the moment DICOM images are pushed via the DICOMweb protocol to an azure DICOM server. Sadly the DSH doesn't have a DICOM server that we can use, so we'll have to use FTPS for 100 days. 

### Documentation

_No response_

### Dependencies

- [x] DSH account for uploading data only for the project UCLH-Foundry/the-rolling-skeleton#77
- [x] de-identification done using #188 

#### Getting hashed identifier

- [API documentation to get a raw tag](https://orthanc.uclouvain.be/api/index.html#tag/Instances/paths/~1instances~1{id}~1content~1{path}/get) Example using demo instance for tag only: <https://orthanc.uclouvain.be/demo/instances/83354841-24928346-008ed987-8032f76c-6363d8eb/content/0010-0020>
- [API documentation for a resource's tags](https://orthanc.uclouvain.be/api/index.html#tag/Instances/paths/~1instances~1{id}~1study/get) : Example to get all common <https://orthanc.uclouvain.be/demo/instances/83354841-24928346-008ed987-8032f76c-6363d8eb/tags?simplify>

#### Downloading dicom data

It [looks like its possible to do this](https://orthanc.uclouvain.be/book/users/rest.html#id16) using the `resourceId`  that we currently use in the dicomweb request in `SendViaStow` in orthanc-anon. 

```
All instances of a study can be retrieved as a zip file as follows:

$ curl http://localhost:8042/studies/6b9e19d9-62094390-5f9ddb01-4a191ae7-9766b715/archive > Study.zip
```

So in python we could do something like this? 

```
# Query orthanc-anon for the study
query = f "{orthanc_url}/studies/{resource_id}/archive"
response_study = requests.get(query, verify=False,   
							  auth=(orthanc_user, orthanc_password))  
if response_study.status_code != 200:  
	raise SpecificException(f"Could not download archive of resource '{resource_id}'")
# get the zip content
zip_content = response_study.content  
```

#### Copying file

@tcouch uses this curl command to send data to the DSH 
```
curl --netrc --tlsv1.2 --ftp-ssl-reqd --ftp-create-dirs -T $ZIPFILE.zip "ftps://filetransfer.idhs.ucl.ac.uk/"
```

We can convert this to python that creates a directory in the FTP server if it doesn't exist. 🎉  [Prototype repo](https://github.com/stefpiatek/docker-ftp-prototype/) that can be used as a basis for automated testing 🎉

milanmlft:
- Adds `upload_dicom_image()` in `pixl_core` to export DICOM images over FTPS to the DSH
- DICOM images are exported as a `.zip` file to `<project_slug>/<pseudo_anon_identifier>.zip`
- Adds a test FTP server Docker container to mock an FTPS server for testing

Fixes #190


pixl_core/src/core/upload.py
```diff
@@ -3,35 +3,47 @@
...
 
 def upload_file(local_file_path: Path) -> str:
+    """Upload local file to hardcoded directory in ftp server."""
+    output_filename = local_file_path.name
+    # Store the file using a binary handler
+    with local_file_path.open("rb") as file_content:
+        upload_content(file_content, output_filename)
+
+
+def upload_content(content: bytes, pseudo_anon_id: str) -> str:
```
stefpiatek, line 25: I think we'd want this to be a generic function which takes in file content and output directory and output filename, rather than doing the database lookup at this point. That way we could reuse it for both writing DICOM files and writing parquet files. Then querying the database for the project happens in a higher level function

pixl_core/src/core/upload.py
```diff
@@ -3,35 +3,47 @@
...
-        ftp.storbinary(command, local_file)
+    command = f"STOR {output_filename}"
+    logger.info("Running %s", command)
+    ftp.storbinary(command, content)
 
     # Close the FTP connection
     ftp.quit()
     logger.info("Done!")
 
+    update_exported_at_and_save(datetime.now)
```
stefpiatek, line 45: and then this would also happen in the higher level function after its written, as currently this function is doing more than uploading a file

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -156,6 +157,34 @@ def SendViaStow(resourceId):
...
 
 
+def SendViaFTPS(resourceId: str) -> None:
+    """
+    Makes a POST API call to upload the resource to a dicom-web server
+    using orthanc credentials as authorisation
+    """
+    ORTHANC_USERNAME = config("ORTHANC_USERNAME")
+    ORTHANC_PASSWORD = config("ORTHANC_PASSWORD")
+
```
stefpiatek, line 167: probably would be good to know this doens't exist when we load up DICOM server rather than at export, so making this in the top level `pixl.py` file might be nicer?

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -156,6 +157,34 @@ def SendViaStow(resourceId):
...
+            msg = "Could not download archive of resource '%s'"
+            raise RuntimeError(msg, resourceId)
+    except requests.exceptions.RequestException:
+        orthanc.LogError(f"Failed to query'{resourceId}'")
+
+    # get the zip content
+    zip_content = response_study.content
+    logging.info("Downloaded data: %s", zip_content)
+
+    upload.upload_content(zip_content, resourceId)
```
stefpiatek, line 185: ooh I think this is the wrong function being called, should it not be `upload_dicom_image`

pixl_core/src/core/_database.py
stefpiatek, line 1: this being called `_database` is unusual, it should be a public module as we're importing it in other modules. Though we already had a `database.py` which doesn't have the connection details in there, which I think is reasonable. Maybe this `_database.py` -> `queries.py`?

pixl_core/src/core/_database.py
```diff
@@ -0,0 +1,76 @@
...
+def get_project_slug_from_db(hashed_value: str) -> str:
+    PixlSession = sessionmaker(engine)
+    with PixlSession() as pixl_session, pixl_session.begin():
+        existing_image = (
+            pixl_session.query(Image)
+            .filter(
+                Image.hashed_identifier == hashed_value,
+            )
+            .one()
+        )
```
stefpiatek, line 46: you could make this into a private method which is also used below as well

pixl_core/src/core/upload.py
```diff
@@ -0,0 +1,119 @@
...
+
+    @sock.setter
+    def sock(self, value: socket) -> None:
+        """When modifying the socket, ensure that it is ssl wrapped."""
+        if value is not None and not isinstance(value, ssl.SSLSocket):
+            value = self.context.wrap_socket(value)
+        self._sock = value
+
+
+def upload_dicom_image(local_file_path: Path, pseudo_anon_id: str) -> None:
```
stefpiatek, line 60: is this used? the pixl extension in dicom anon uses `upload_content` and I guess it should be using this function instead

pixl_core/src/core/upload.py
```diff
@@ -0,0 +1,119 @@
...
+            file_content,
+            remote_dir=remote_directory,
+            remote_file=f"{pseudo_anon_id}.zip",
+        )
+
+    update_exported_at_and_save(pseudo_anon_id, datetime.now(tz=timezone.utc))
+    return output_path
+
+
+def upload_content(content: BinaryIO, *, remote_dir: str, remote_file: str) -> str:
```
stefpiatek, line 79: probably worth making this a private method so that we know we shouldn't be calling it directly

pixl_core/tests/conftest.py
```diff
@@ -13,6 +13,157 @@
...
+from pathlib import Path
 
-os.environ["ENV"] = "test"
+import pytest
+from core.database import Base, Extract, Image
+from sqlalchemy import Engine, create_engine
+from sqlalchemy.orm import Session, sessionmaker
+
+os.environ["RABBITMQ_USERNAME"] = "guest"
+os.environ["RABBITMQ_PASSWORD"] = "guest"  # noqa: S105 Hardcoding password
```
stefpiatek, line 28: might be worth just adding the `S105` exception into the top level ruff exclusions for anything `**tests**`

pixl_core/tests/conftest.py
```diff
@@ -13,6 +13,157 @@
...
+os.environ["FTP_USER_NAME"] = "pixl"
+os.environ["FTP_USER_PASS"] = "longpassword"  # noqa: S105 Hardcoding password
+os.environ["FTP_PORT"] = "20021"
+
+TEST_DIR = Path(__file__).parent
+STUDY_DATE = datetime.date.fromisoformat("2023-01-01")
+
+
+@pytest.fixture(scope="package")
+def run_containers() -> None:
```
stefpiatek, line 41: strictly speaking this returns `subprocess.CompletedProcess[str]`

pixl_core/tests/conftest.py
```diff
@@ -13,6 +13,157 @@
...
+
+
+@pytest.fixture(scope="package")
+def data() -> Path:
+    """Directory containing the test data for uploading to the ftp server."""
+    return TEST_DIR / "data"
+
+
+@pytest.fixture(scope="package")
+# make mounted_data depend on run_containers; the teardown needs the Docker containers to be down
```
stefpiatek, line 66: this should probably be in the docstring or a comment within the function

pixl_core/tests/conftest.py
```diff
@@ -13,6 +13,157 @@
...
+        b"docker compose down --volumes",
+        check=True,
+        cwd=TEST_DIR,
+        shell=True,  # noqa: S602
+        timeout=60,
+    )
+
+
+@pytest.fixture(scope="package")
+def data() -> Path:
```
stefpiatek, line 60: data is quite generic, maybe `test_dir`, `data_dir` or `test_data`

pixl_core/tests/conftest.py
```diff
@@ -13,6 +13,157 @@
...
+    """Directory containing the test data for uploading to the ftp server."""
+    return TEST_DIR / "data"
+
+
+@pytest.fixture(scope="package")
```
stefpiatek, line 65: do we want teardown to happen after each function rather than package, so that if two tests wrote to the same place that we'd be sure that each of them wasn't interacting?

pixl_core/tests/test_upload.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@pytest.mark.usefixtures("run_containers")
+def test_upload_content(data, mounted_data) -> None:
+    """Tests that file is present on the endpoint after upload"""
+    local_file = data / "public.zip"
+    with local_file.open("rb") as handle:
+        upload_content(handle, remote_file="public.zip", remote_dir="new_dir")
+
+    assert (mounted_data / "new_dir" / "public.zip").exists()
```
stefpiatek, line 32: probably can remove this test as we've got that covered in the upload dicom image?

stefpiatek (commented):
Looking good. Need to make some changes for it to work in production

orthanc/orthanc-anon/plugin/pixl.py
ruaridhg: Wouldn't the zip_content be a BinaryIO in this case so an extra argument of remote_dir would need to be added? Or need to pass in a file_path instead of zip_content so would need to download first

pixl_core/src/core/_database.py
ruaridhg: Ah because it's in core, there should only be public modules?

orthanc/orthanc-anon/plugin/pixl.py
milanmlft: Yeah good point, I think we can just merge `upload_dicom_image()` and `upload_content()`, because for the parquet files we'll need a different function anyway.

pixl_core/tests/conftest.py
milanmlft: Ah yeah, this was due to my repeated attempts to get around file permission errors in Docker volumes on GHA, but ended up "fixing" it by just `ignore_errors=True` when removing files 🙃 
So don't really need this anymore, will fix

pixl_core/src/core/_database.py
milanmlft: Yeah, I think we followed the same pattern as in `pixl_dcmd._database.py`

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -156,6 +155,29 @@ def SendViaStow(resourceId):
...
+        success_code = 200
+        if response_study.status_code != success_code:
+            msg = "Could not download archive of resource '%s'"
+            raise RuntimeError(msg, resourceId)
+    except requests.exceptions.RequestException:
+        orthanc.LogError(f"Failed to query'{resourceId}'")
+
+    # get the zip content
+    zip_content = response_study.content
+    logging.info("Downloaded data: %s", zip_content)
```
stefpiatek, line 176: do we want to log the zip content? This'll be binary data right? Happy for you to log the resourceId

stefpiatek (approved):
Thanks for getting this in 🎉 

Happy for you to merge, some small suggestions remaining I think

PR #229, opened 2024-01-19
Debug and get working

stefpiatek:
Turns out I should take my own advice of [RTFM](https://github.com/gimoh/docker-pureftpd#passive-ftp-ports):

> You also need to choose a range of ports for data channels, say you want to support 10 concurrent users, pick 10 ports like 30000-30009 and pass that to docker run as --publish option as well as -p (passive port range) to pureftpd.

🤦 When it was all running in the docker network this wasn't a problem because the ports would just be there

Also had some tidying up and making the file upload method a bit easier to use when we've already got binary data (but not a file to read)

### DIFF
--- a/pixl_core/src/core/upload.py
+++ b/pixl_core/src/core/upload.py
@@ -4,15 +4,15 @@ import ftplib
 import logging
 import os
 from ftplib import FTP_TLS
-from pathlib import Path
+from typing import BinaryIO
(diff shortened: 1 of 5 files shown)


stefpiatek:
Doesn't need to be used as I think we've got something which runs on apple silicone and intel chips

PR #230, opened 2024-01-19
Use FTP_TLS for testing

stefpiatek:
Testing out FTP upload with DSH and found some implementation details we missed for the FTP connection. This solved our problem with running the test server and was able to reconfigure the ftp-server container to be used 🎉 

Also use datetime utc so don't need to install extra library


ruaridhg (approved):
LGTM and it worked locally for me!

PR #246, opened 2024-01-29
Remove the `test.csv` file as it's not needed anymore

milanmlft:
Lingering artefact from when we moved to Parquet files in #186


ruaridhg (approved):
👍

PR #251, opened 2024-01-29
Fix docstring of `ImagingStudy` class

milanmlft:
Artefact from copying over EHR API docstrings


stefpiatek (approved):
Thanks for fixing this up

PR #264, opened 2024-01-29
Export current extract parquet files over FTPS

### LINKED ISSUE #192: Export current extract parquet files to DSH

The public parquet files are zipped and copied via FTPS to the DSH in a directory for the extract name slug

### Testing

Use the same testing setup for #190, but instead testing the parquet files 

### Documentation

_No response_

### Dependencies

After #190 

### Details and Comments

Could probably do a manual upload to the DSH instead if required.
_Might_ be a nice option to merge the two export commands.

ruaridhg:
Closes #192 
Changes:

- Change SendViaStow to SendViaFTPS in orthanc-anon
- upload_parquet_files added to core module
- Tear down test data through docker otherwise later tests fail
- Tests for successful upload and exception thrown in nothing to export
- app in pixl_ehr to SendViaFTPS


orthanc/orthanc-anon/plugin/pixl.py
milanmlft, line 1: Any reason why you touched this? This will conflict with #268 🥲

pixl_ehr/src/pixl_ehr/main.py
```diff
@@ -96,6 +97,16 @@ def export_radiology_as_parquet(export_params: ExportRadiologyData) -> None:
...
 
 
+@app.post(
+    "/send-parquet-files-via-FTPS",
+    summary="Copy/send all ParquetExports via FTPS",
+)
+def SendViaFTPS(project_name: str, extract_datetime: datetime) -> None:
+    """Send parquet files via FTPS"""
+    pe = ParquetExport(project_name, extract_datetime)
+    upload_parquet_files(pe)
```
milanmlft, line 107: Hmm I would suggest renaming to `send_via_ftps()` to be consistent and to avoid confusion with the callback in the Orthanc plugin.

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -202,7 +202,7 @@ def OnChange(changeType, level, resource) -> None:  # noqa: ARG001
 
     if changeType == orthanc.ChangeType.STABLE_STUDY and ShouldAutoRoute():
         print("Stable study: %s" % resource)  # noqa: T201
-        SendViaStow(resource)
```
milanmlft, line 205: Any reason why you touched this? This will conflict with #268 🥲

pixl_core/tests/conftest.py
```diff
@@ -66,18 +67,21 @@ def test_zip_content() -> BinaryIO:
...
     """
     yield TEST_DIR / "ftp-server" / "mounts" / "data"
-    sub_dirs = [
-        f.path for f in os.scandir(TEST_DIR / "ftp-server" / "mounts" / "data") if f.is_dir()
-    ]
     # Tear down the directory after tests
-    for sub_dir in sub_dirs:
-        shutil.rmtree(sub_dir, ignore_errors=True)
+    subprocess.run(
+        b"docker compose exec ftp-server sh -c 'rm -rf /home/pixl/*'",
```
stefpiatek, line 79: Does this work?
```suggestion
        b"docker compose exec ftp-server sh -c 'rm -r /home/pixl/*'",
```

pixl_ehr/src/pixl_ehr/main.py
stefpiatek: Yeah agreed, as its not an orthanc plugin we should be snake case.

I'd also be tempted to just have a single `export-patient-data` API endpoint which writes the radiology report and exports it via FTPS (calling a`send_via_ftps` function so that we can keep the same tests as we've got currently)

stefpiatek (approved):
Couple of suggestions 🎉

orthanc/orthanc-anon/plugin/pixl.py
ruaridhg: I see you've now changed your code to match?

pixl_core/tests/conftest.py
ruaridhg: Yes it does

orthanc/orthanc-anon/plugin/pixl.py
milanmlft: No we made the change already before, which is why it's going to cause a conflict now 🙃 
I don't think this is relevant to the parquet files exporting, so shouldn't be changed here (ideally we would have done this in #226 but I guess we forgot)

stefpiatek: Seems like an easy confict to resolve, not sure why its a problem?

milanmlft (changes requested):
I fixed the problem with the `FTP_HOST` key not being found. I think what happened is that by moving the FTP uploading functionality into `export_patient_data()`, the unit tests would try to run this but then fail because it can't find that environment variable. Further on that, it also needs an actual FTP server to be up and running, so added that to the `docker-compose.yml` as well (though this will need to be updated after merging #268).

The CI is currently still failing because of some missing files, so I think there is still a wrongly configured file path somewhere.
That being said, I don't see any actual unit tests that check wether the parquet files have been uploaded to the FTP server, like we're doing in `pixl_core/tests/test_upload.py` for the DICOM images. So probably still need to add that?

stefpiatek:
>  That being said, I don't see any actual unit tests that check wether the parquet files have been uploaded to the FTP server, like we're doing in pixl_core/tests/test_upload.py for the DICOM images. So probably still need to add that?

Are they not also in `pixl_core/tests/test_upload.py`?

ruaridhg:
> @milanmlft
> 
> > That being said, I don't see any actual unit tests that check wether the parquet files have been uploaded to the FTP server, like we're doing in pixl_core/tests/test_upload.py for the DICOM images. So probably still need to add that?
> 
> Are they not also in `pixl_core/tests/test_upload.py` `test_upload_parquet`?
> 
> I'm confused though, why are we testing in ehr, when all of this functionality that we care about is in core (which already covers the test)?

Yeah, the unit tests are in `test_upload_parquet`, but since test_processing in ehr tests the function that I modified to include uploading parquet it now fails so have been trying to fix ehr test errors

milanmlft:
> Are they not also in `pixl_core/tests/test_upload.py` `test_upload_parquet`?
> 
> I'm confused though, why are we testing in ehr, when all of this functionality that we care about is in core (which already covers the test)?

We're not actually testing in `ehr`, but we're testing `export_patient_data()`, which now includes a call to `send_via_ftps()`, which fails when there is no FTPS server available.

I agree with keeping the FTPS-related testing in core. But then we should probably monkeypatch the `ehr` tests so that  it doesn't try to connect to an FTPS server (and then we can remove the FTPS setup in `ehr` again).

stefpiatek:
> I agree with keeping the FTPS-related testing in core. But then we should probably monkeypatch the ehr tests so that it doesn't try to connect to an FTPS server (and then we can remove the FTPS setup in ehr again).

👌 second option is to create a separte  function `export_reports_to_parquet` and test that instead of testing the top level API call. That feels like less faffing around but I'm easy

pixl_ehr/tests/docker-compose.yml
```diff
@@ -65,3 +65,28 @@ services:
...
+    environment:
+      ADDRESS: "localhost"
+      USERS: pixl|longpassword|/home/pixl
+      TLS_KEY: /etc/ssl/private/localhost.key
+      TLS_CERT: /etc/ssl/private/localhost.crt
+    healthcheck:
+      test: ping localhost:21 -w 2
+      interval: 10s
+      timeout: 5s
+      retries: 5
```
milanmlft, line 92:
```suggestion
```

Can be removed again once we avoid calling `send_via_ftps()` during tests

milanmlft (dismissed):
Think it's looking good now! Just the `osimis/orthanc` docker image seems to have some problems upstream. Just re-running the CI might work 🤷‍♂️

pixl_ehr/tests/test_processing.py
```diff
@@ -346,7 +346,6 @@ async def test_radiology_export_multiple_projects(example_messages, tmp_path) ->
         await process_message(mess)
 
     # ACT
```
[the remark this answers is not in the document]
ruaridhg: > 👌 second option is to create a separte function `export_reports_to_parquet` and test that instead of testing the REST API call. That feels like less faffing around but I'm easy

I assumed this is what you meant by having `export_radiology_as_parquet` and `send_via_ftps` as separate functions called in a single API endpoint i.e. `export_patient_data` but testing just `export_radiology_as_parquet` in `test_processing` since send via ftps is tested elsewhere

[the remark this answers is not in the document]
milanmlft: System test is failing now because of the call to `export_patient_data()` in `pixl export-patient-data` which in turn calls `send_via_ftps()` (which we want I think!). However, this branch doesn't have an FTPS server set up yet for the system tests (that's under way in #268), so it's failing because of the same reason as before.

I think we can leave this for now and wait until #268 is fixed and merged and then merge `main` into this branch and add an additional check in the system tests to verify the successful upload of the parquet files.

milanmlft:
System test is failing now because of the call to `export_patient_data()` in `pixl export-patient-data` which in turn calls `send_via_ftps()` (which we want I think!). However, this branch doesn't have an FTPS server set up yet for the system tests (that's under way in #268), so it's failing because of the same reason as before.

I think we can leave this for now and wait until #268 is fixed and merged and then merge `main` into this branch and add an additional check in the system tests to verify the successful upload of the parquet files.

stefpiatek (commented):
Looking good! Can you add in a check for the parquet files existing in the system test?

milanmlft:
Added a check to the system test to verify uploading of Parquet files
The system test will fail currently, **because we're missing the `radiology.parquet` file**. Don't we want that file to get uploaded to the DSH as well? 🤔

stefpiatek (approved):
🎉

milanmlft (approved):

Issue #195, opened 2023-11-17
PIXL pipeline: Update python version to 3.12


stefpiatek:
### Definition of Done / Acceptance Criteria

To maintain [python support](https://endoflife.date/python) during life of PIXL project.

PIXL modules currently running on python 3.9 and 3.10. Should update these to 3.12 where we can. 

Sadly orthanc seems to be always have Docker containers with 3.9 so may need to accept that for now in `pixl_dcmd`

### Testing

Tests pass, and carry out a test run on the GAE

- [x] `core`
- [x] `cli`
- [x] `ehr-api`
- [x] `imaging-api`
- ~`pixl_dcmd`~ (Orthanc still requires 3.9)
- [x] `hasher`
- [x] `pytest-pixl`
- [x] `system-test`
- [x] Test run on GAE

### Details and Comments

- [x] Update `pydantic` to latest version and replace `validator` with `field_validator`

As part of #309, we're using `pydantic.validator`, however in `pydantic>=2.0` the recommended method is to use `pydantic.field_validator`.

milanmlft:
`PyYAML 6.0` gives installation errors on 3.12.
Do we still need `PyYAML` though? If yes, should increase version requirement to `6.0.1`

stefpiatek:
I think yeah we do need it actually, let's update to 6.0.1

stefpiatek:
[orthancteam/orthanc](https://hub.docker.com/r/orthancteam/orthanc) has been documented as the successor to [osimis/orthanc](https://hub.docker.com/r/osimis/orthanc) which is no longer being updated. Still python 3.9.2 but hopefully that should mean there'll be newer releases soon

milanmlft:
Confirmation that Orthanc should move to Python 3.11 soon: https://github.com/orthanc-server/orthanc-builder/issues/19#issuecomment-1976938758

milanmlft:
Orthanc containers have been updated to Python 3.11! 🎉 
https://github.com/orthanc-server/orthanc-builder/blob/master/release-notes-docker-images.md#2433

PR #267, opened 2024-01-30
Bump required version of PyYAML to 6.0.1 for cli, pixl_dcmd and pixl_ehr

jstutters:
This allows installation with Python 3.12 without breaking under 3.9.  Addresses #195.


milanmlft (approved):
🚀

PR #268, opened 2024-01-30
Add FTPS server to system test and test successful upload

### LINKED ISSUE #240: Use FTPS server in system tests

- [x] FTPS docker container defined in `/test/dummy-services`
- [x] DICOM FTPS upload tested in system tests
- [ ] parquet FTPS upload tested in system tests
  - probably worth changing the current `ehr-api` endpoint for writing reports parquet to instead `upload-via-ftps` which saves the parquet file to the right directory and then uploads the reports and OMOP parquet files  

### Dependencies

- [ ] #192

milanmlft:
- Some file reorganisation: move cogstack dummy service to `test/dummy-services/cogstack`
- Move the test `ftp-server` from `pixl_core` to `test/dummy-services/ftp-server` 
- Make sure `orthanc-anon` uploads images over FTPS after anonymisation
- Add check to system test to verify a `zip` file has been uploaded to FTPS server
- [x] Cleanup ftp server files after tests

Fixes #240


.env.sample
```diff
@@ -43,16 +43,16 @@ ORTHANC_RAW_MAXIMUM_STORAGE_SIZE= // MB
 
 # PIXL Orthanc anon instance
 ORTHANC_ANON_USERNAME=
-ORTHANC_ANON_PASSWORD=     
+ORTHANC_ANON_PASSWORD=
 ORTHANC_ANON_AE_TITLE=
 ORTHANC_ANON_HTTP_TIMEOUT=60
-ORTHANC_AUTOROUTE_ANON_TO_AZURE=false
+ORTHANC_AUTOROUTE_ANON_TO_FTPS=true
```
stefpiatek, line 49: We want to actively choose to export data so I think false is a better default to minimise accidental data transfer during development

Maybe `ORTHANC_AUTOROUTE_ANON_TO_ENDPOINT` because we will choose to route to azure/DSH/another DICOMWeb endpoint

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -180,17 +184,17 @@ def SendViaFTPS(resourceId: str) -> None:
...
-    Checks whether ORTHANC_AUTOROUTE_ANON_TO_AZURE environment variable is
+    Checks whether ORTHANC_AUTOROUTE_ANON_TO_FTPS environment variable is
     set to true or false
     """
-    return os.environ.get("ORTHANC_AUTOROUTE_ANON_TO_AZURE", "false").lower() == "true"
+    return os.environ.get("ORTHANC_AUTOROUTE_ANON_TO_FTPS", "false").lower() == "true"
 
 
-def OnChange(changeType, level, resource) -> None:  # noqa: ARG001
+def OnChange(changeType: str, _level: None, resource: str) -> Any:
```
stefpiatek, line 193: I'm guessing there's a reason with the orthanc plugin system that means we've changed these to returning `Any`?

stefpiatek (approved):
Looks good to me, suggestion and a question

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -180,17 +184,17 @@ def SendViaFTPS(resourceId: str) -> None:
...
     """
     Three ChangeTypes included in this function:
     - If a study if stable and if ShouldAutoRoute returns true
```
peshence, line 196:
```suggestion
    - If a study is stable and if ShouldAutoRoute returns true
```

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -202,7 +206,7 @@ def OnChange(changeType, level, resource) -> None:  # noqa: ARG001
 
     if changeType == orthanc.ChangeType.STABLE_STUDY and ShouldAutoRoute():
         print("Stable study: %s" % resource)  # noqa: T201
-        SendViaStow(resource)
+        SendViaFTPS(resource)
```
peshence, line 209: should we delete the SendViaStow function as it's unused?

.env.sample
ruaridhg: Had a question about this during onboarding, but assuming this is set as default false how do we actually trigger the data to be sent from anon to FTPS currently?

milanmlft: I assume by setting it locally (e.g. in `test/.env.test`) to `true`?

orthanc/orthanc-anon/plugin/pixl.py
milanmlft: Tbf, I just added these to get rid of the overload of squiggly lines in my VSCode 😅 
There weren't any type annotations before and we're not enforcing them in this file.
(I guess I could also just ignore that file in my VSCode...)

orthanc/orthanc-anon/plugin/pixl.py
milanmlft: I would say yes, but maybe it will become relevant again at some piont? @stefpiatek?

jeremyestein (dismissed):

test/scripts/check_ftps_upload.py
```diff
@@ -0,0 +1,33 @@
...
+logger = logging.getLogger(__name__)
+
+PARQUET_PATH = Path(__file__).parents[1] / "resources" / "omop"
+MOUNTED_DATA_DIR = Path(__file__).parents[1] / "dummy-services" / "ftp-server" / "mounts" / "data"
+
+project_name, omop_es_timestamp = config_from_log_file(PARQUET_PATH)
+expected_output_dir = MOUNTED_DATA_DIR / project_name
+
+# Test whether DICOM images have been uploaded
+glob_list = [expected_output_dir.glob("*.zip")]
```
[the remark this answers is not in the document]
milanmlft: Whoops yes exactly! Thank you!

test/scripts/check_ftps_upload.py
```diff
@@ -0,0 +1,33 @@
...
+logger.info(f"glob_list: {glob_list}")
+
+assert len(glob_list) == 1
+
+# TODO: check parquet files upload
```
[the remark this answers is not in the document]
milanmlft: Yeah so that would be part of #264. I think we can merge this PR in first and then complete the system test in #264.

pixl_core/tests/conftest.py
```diff
@@ -71,9 +72,11 @@ def mounted_data() -> Path:
     The mounted data directory for the ftp server.
     This will contain the data after successful upload.
     """
-    yield TEST_DIR / "ftp-server" / "mounts" / "data"
+    yield DUMMY_SERVICES_DIR / "ftp-server" / "mounts" / "data"
     sub_dirs = [
-        f.path for f in os.scandir(TEST_DIR / "ftp-server" / "mounts" / "data") if f.is_dir()
+        f.path
+        for f in os.scandir(DUMMY_SERVICES_DIR / "ftp-server" / "mounts" / "data")
```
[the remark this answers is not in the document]
milanmlft: WDYM?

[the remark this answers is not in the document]
milanmlft: I assume you mean the repeated code on lines `75` and `78`?

pixl_core/tests/conftest.py
```diff
@@ -71,9 +72,11 @@ def mounted_data() -> Path:
...
+        if f.is_dir()
```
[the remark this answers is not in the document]
milanmlft: This was added in #226 where I assumed there would only ever be directories created here. That being said, I think this teardown has been updated in #264, so can probably leave this here.

pixl_core/tests/conftest.py
```diff
@@ -35,6 +35,7 @@
 os.environ["FTP_PORT"] = "20021"
 
 TEST_DIR = Path(__file__).parent
+DUMMY_SERVICES_DIR = Path(__file__).parents[2] / "test" / "dummy-services"
```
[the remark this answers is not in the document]
milanmlft: Hmm yeah, but we need it to point to the volume that is mounted in the `ftp-server` container, which is at `test/dummy-services/ftp-server/mounts/data`, so a `tmp_path` wouldn't work in this case?

.env.sample
stefpiatek: > Had a question about this during onboarding, but assuming this is set as default false how do we actually trigger the data to be sent from anon to FTPS currently?

and likewise in production we change the .env to `true`

orthanc/orthanc-anon/plugin/pixl.py
stefpiatek: yeah we want to keep that functionality when we start stopping this from being hardcoded and make the endpoint we send to configurable

pixl_core/tests/conftest.py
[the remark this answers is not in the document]
stefpiatek: you don't pip install tests though, so this will always be relative to the `conftest.py`, unless I'm misunderstanding

.env.sample
ruaridhg: Okay great, just wanted to clarify thanks

pixl_core/tests/conftest.py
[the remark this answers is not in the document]
jeremyestein: Erm, yes, sorry

pixl_core/tests/conftest.py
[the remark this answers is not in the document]
jeremyestein: Ah, it's possible I only saw it in non-test code, where we use it like this:
`HOST_EXPORT_ROOT_DIR = Path(__file__).parents[3] / "exports"`
which might be a problem there, if not here.
Not for this PR perhaps, but I think the paths should be outside the source tree and configurable.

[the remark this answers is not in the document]
milanmlft: Alright, makes sense. But so it's fine to leave it like this for this PR?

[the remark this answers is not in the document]
stefpiatek: Yeah I think so, practially we're not expecting to put this on pypi so will be the source location. Happy to make an issue to ensure that we're doing this in the long run

jstutters:
#272 may help with this by fixing a problem in pixl_dcmd that was preventing DICOMs received by orthanc-anon from being stored.

pixl_core/tests/conftest.py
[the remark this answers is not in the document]
milanmlft: @jeremyestein Could you please approve then so we can merge (once the system test has been fixed)? 🙏

.env.sample
```diff
@@ -83,3 +83,8 @@ RABBITMQ_PASSWORD=
 # PACS extraction API
 PIXL_DICOM_TRANSFER_TIMEOUT=240
 PIXL_QUERY_TIMEOUT=10
+
+# FTP server
+FTP_HOST=
+FTP_USER_NAME=
+FTP_USER_PASS=
```
stefpiatek, line 90: Thinking about it, probably should be `FTP_USER_PASSWORD` for consistency across all our other envs

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -160,37 +162,59 @@ def SendViaFTPS(resourceId: str) -> None:
...
+    Queries the Orthanc instance to get the PatientID for a given resource.
+    When anonymisation has been applied, the PatientID is the pseudo-anonymised ID.
+    """
+    query = f"{ORTHANC_URL}/studies/{resourceId}"
+    fail_msg = "Could not query study for resource '%s'"
+    response_study = _make_query(resourceId, query, fail_msg)
+    return json.loads(response_study.content.decode())["PatientMainDicomTags"]["PatientID"]
+
+
+def _make_query(resourceId: str, query: str, fail_msg: str) -> requests.Response:
```
stefpiatek, line 189: Might be simpler to just call it query
```suggestion
def _query(resourceId: str, query: str, fail_msg: str) -> requests.Response:
```

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -160,37 +162,59 @@ def SendViaFTPS(resourceId: str) -> None:
...
+
+    # get the zip content
+    zip_content = response_study.content
+    logging.info("Downloaded data for resource %s", resourceId)
+
+    upload.upload_dicom_image(zip_content, GetPatientID(resourceId))
+    logging.info("Uploaded data to FTPS for resource %s", resourceId)
+
+
+def GetPatientID(resourceId: str) -> str:
```
stefpiatek, line 178: why are we CamelCasing this?

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -160,37 +162,59 @@ def SendViaFTPS(resourceId: str) -> None:
...
+    """
+    Queries the Orthanc instance to get the PatientID for a given resource.
+    When anonymisation has been applied, the PatientID is the pseudo-anonymised ID.
+    """
+    query = f"{ORTHANC_URL}/studies/{resourceId}"
```
stefpiatek, line 183: I think its cleaner to use the suggestion from #190 `{dicom-server}/instances/{resource-id}/content/0010-0020` so we just query the tag directly rather than downloading all of the dicom tags

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -160,37 +162,59 @@ def SendViaFTPS(resourceId: str) -> None:
...
     """
-    Checks whether ORTHANC_AUTOROUTE_ANON_TO_AZURE environment variable is
+    Checks whether ORTHANC_AUTOROUTE_ANON_TO_ENDPOINT environment variable is
     set to true or false
     """
-    return os.environ.get("ORTHANC_AUTOROUTE_ANON_TO_AZURE", "false").lower() == "true"
+    return os.environ.get("ORTHANC_AUTOROUTE_ANON_TO_ENDPOINT", "false").lower() == "true"
+
+
+def _AzureAvailable() -> bool:
```
stefpiatek, line 208: also should prefer snake case for anything that isn't an Orthanc plugin IMHO

stefpiatek (approved):
🎉 Some simplification can be done, happy for you to merge once you're happy

orthanc/orthanc-anon/plugin/pixl.py
milanmlft: Ah yes, guess I was just following the style from the file, but makes sense to put this in snake_case and maybe make it private with as `_get_patient_id`?

stefpiatek: Yeah its a bit of a style call but I think anything that isn't directly used by orthanc should follow pep8

orthanc/orthanc-anon/plugin/pixl.py
milanmlft: Aaah but this doesn't actually work with the `resourceID` 😭 
AFAIK `resourceID` can only be used for `/studies/` queries 😕

milanmlft: I've looked into it a bit more and I don't see a way to get to the specific tag that isn't more cumbersome then the current query I'm doing, so think I will stick with that.

milanmlft:
System test still failing due to a `[Errno 111] Connection refused` error when running the `STOR` command to upload the zip file to FTPS. What is weird is that logging into the FTPS server works fine and even creating and changing working directories. But then commands such as `ftp.dir() and `ftp.storbinary()` fail with the mentioned error.

This all runs fine when running the unit test in `pixl_core`, which uses a similar setup. The main differences are:
- Running the `ftp-server` in a docker compose network
- Running the FTP uploading functionality from within the `orthanc-anon` container, which is on **Python 3.9**

Issue #212, opened 2024-01-08
Configuration to filter omop data by procedure concept


stefpiatek:
### Definition of Done / Acceptance Criteria

When reading data from OMOP ES, there is a way to define which omop `procedure_concept_id`s are used for imaging procedures. 

### Details and Comments

What would be really nice is if this was part of the extract log file. Failing that we could have this as a CLI argument perhaps?

peshence:
@stefpiatek this is public/PROCEDURE_OCCURRENCE.parquet
```json
{"procedure_occurrence_id":1,"person_id":1,"procedure_concept_id":4200610,"procedure_date":"2021-07-01T00:00:00.000Z","procedure_datetime":1625127300000,"procedure_type_concept_id":32817,"procedure_end_date":null,"procedure_end_datetime":null,"modifier_concept_id":0,"quantity":null,"visit_occurrence_id":null,"procedure_source_value":null,"procedure_source_concept_id":0,"modifier_source_value":null}
{"procedure_occurrence_id":2,"person_id":1,"procedure_concept_id":4058335,"procedure_date":"2021-07-01T00:00:00.000Z","procedure_datetime":1625140800000,"procedure_type_concept_id":32817,"procedure_end_date":null,"procedure_end_datetime":null,"modifier_concept_id":0,"quantity":null,"visit_occurrence_id":null,"procedure_source_value":null,"procedure_source_concept_id":0,"modifier_source_value":null}
{"procedure_occurrence_id":3,"person_id":2,"procedure_concept_id":4327032,"procedure_date":"2020-05-01T00:00:00.000Z","procedure_datetime":1588339813000,"procedure_type_concept_id":32817,"procedure_end_date":null,"procedure_end_datetime":null,"modifier_concept_id":0,"quantity":null,"visit_occurrence_id":null,"procedure_source_value":null,"procedure_source_concept_id":0,"modifier_source_value":null}
{"procedure_occurrence_id":4,"person_id":2,"procedure_concept_id":4163872,"procedure_date":"2020-05-23T00:00:00.000Z","procedure_datetime":1590240671000,"procedure_type_concept_id":32817,"procedure_end_date":null,"procedure_end_datetime":null,"modifier_concept_id":0,"quantity":null,"visit_occurrence_id":null,"procedure_source_value":null,"procedure_source_concept_id":0,"modifier_source_value":null}
[...]

PR #271, opened 2024-01-31
 Ensure only imaging studies are processed from OMOP ES parquet files #212 

peshence:
- filter on accession number not none
- remove accession number from test parquet files for all but expected two xray images

Talked about switching to e.g. csv but pandas was struggling with converting datetimes and I thought it was more trouble than ti was worth


stefpiatek (approved):
Yeah that seems fair, I think something like json/toml might be easier but also happy with this. Looks like system tests are still failing CI

PR #272, opened 2024-01-31
More accurate orthanc-anon system test

jstutters:
Fixes some problems with the DICOM anonymisation process in the test environment (missing envvars, persisting of hashes to DB) and ensures that orthanc-anon has received and stored the test DICOMs.


jstutters:
system-test runs to completion outside Actions so speculative commits incoming.

Issue #237, opened 2024-01-26
Reduce reliance on docker containers for module tests


stefpiatek:
### Definition of Done / Acceptance Criteria

Use mocking and monkeypatching to reduce the need to spin up docker containers for modules

### Details and Comments

`pixl_core` will need to have the FTPS server to test that it can write to the server, but otherwise investigate which services are not needed

jeremyestein:
For the sake of those containers that are still needed for testing, we could look into speeding up the image builds.
The usual trick of installing your requirements separately doesn't work for us because we're using pyproject.toml files:
```
COPY requirements.txt .
RUN pip install -r requirements.txt
COPY . .
RUN pip install
```

So we end up with some very long running `pip install` commands that download all the dependencies and re-install them when only our code has changed.

jstutters:
[pip-tools](https://github.com/jazzband/pip-tools) can give you `requirements.txt` from a `pyproject.toml` but then you have the headache of keeping those in sync.  Run the tests from a Makefile?
```
python -m piptools compile -o requirements.txt pyproject.toml
```

stefpiatek:
> [pip-tools](https://github.com/jazzband/pip-tools) can give you requirements.txt from a pyproject.toml but then you have the headache of keeping those in sync. Run the tests from a Makefile?

Yeah as this is all to go with steps in dockerfiles, can just have that as an earlier step and then install the dependencies before we install the package

peshence:
Me and @ruaridhg had a look at the CLI module, where the only tests needing the rabbitmq docker are in [test_queue_start_and_stop_parquet.py](https://github.com/UCLH-Foundry/PIXL/blob/main/cli/tests/test_queue_start_and_stop_parquet.py). We started to mock out the consumer and producer but found ourselves wondering if we're actually testing anything then? I would argue all the tests there are unnecessary

peshence:
Looking for input, details below, but basic questions are:
- do we value functional tests of each module? e.g. `pixl_core/tests/patient_queue/{producer/subscriber}.py` assume yes as there is little logic in the underlying code but we need to have some sort of handle on its condition.
- In order to test a lot of our code we need running services. Instead of bringing up everything each time, we could simply run within a dev container with all the services running. This would present some load onto developers' systems. Is everybody up for it? Do you have alternative suggestions?

## CLI
only needs a rabbitmq instance for the `cli/tests/test_queue_start_and_stop_parquet.py` tests which try to:
[...]

PR #276, opened 2024-02-01
#237 Reduce docker reliance in PIXL-EHR

peshence:
- remove unused rabbitmq instance
- add docker compose to testconf
- remove run_tests scripts

Still needs docker because it tests whether something has happened in the DB. Not familiar enough with things to tell if that's necessary. 

But if it is, perhaps we could make a dev container with all the test infrastructure running locally? That way we don't have to keep turning docker on and off for each test run


pixl_ehr/tests/docker-compose.yml
```diff
@@ -58,10 +46,3 @@ services:
         N_TABLE_ROWS: 2
     ports:
       - "35433:5432"
-
-    healthcheck:
-      test: ["CMD", "pg_isready", "-U", "postgres"]
-      start_period: 30s
-      interval: 10s
-      timeout: 30s
-      retries: 5
```
milanmlft, line 67: Why remove only the healthcheck? If we keep the postgres database, would probably still need this no?

milanmlft:
Nice! I think regarding the PIXL postgres database, we might also be able to mock it instead of having to spin up the Docker service? But not sure how feasible that is.
I definitely like the idea of a devcontainer.

peshence:
> Nice! I think regarding the PIXL postgres database, we might also be able to mock it instead of having to spin up the Docker service? But not sure how feasible that is. I definitely like the idea of a devcontainer.

I would generally agree about mocking when it comes to unit tests. However, these are not unit tests - they are functional black box tests, in the sense that they don't test specific pieces of logic but rather the function of the module (and really its communication with other modules, dbs, etc).

stefpiatek (approved):
Nice, for now agree keeping the postgres in is fine. Ultimately we'll aim to not interact with postgres at all and this will be an exporting api

PR #277, opened 2024-02-05
Update PR template for VOXL milestone

stefpiatek:
- Update template to be for VOXL (though maybe this naming will change?)
- use checklists under headings to try and make checklists a default choice when filling


HChughtai (approved):
👍🏽

PR #281, opened 2024-02-06
fix[core]: async test actually running now

peshence:
as mentioned [here](https://github.com/UCLH-Foundry/PIXL/issues/237#issuecomment-1927517039) test on async consumer in core was not being awaited and always passed.

Simplified and made sure it **does** run now.


peshence:
I haven't the foggiest how me changing a unit test would break the system tests...

stefpiatek:
> I haven't the foggiest how me changing a unit test would break the system tests...

Ah this is the flakiness of the ftps server, which I think @milanmlft is going to work on to use the python server as you suggested

pixl_core/tests/conftest.py
```diff
@@ -53,7 +53,7 @@ def run_containers() -> subprocess.CompletedProcess[bytes]:
         timeout=60,
     )
     subprocess.run(
-        b"docker compose down --volumes",
+        b"docker compose down",
```
stefpiatek, line 56: Ah intersting, why have we removed this?

pixl_core/tests/patient_queue/test_subscriber.py
```diff
@@ -37,51 +33,29 @@
...
-            self.get_async_result(pc.run(callback=consume))
-
-        assert counter == 1
+@pytest.mark.asyncio()
+@pytest.mark.usefixtures("run_containers")
+async def test_create() -> None:
+    """Checks consume is working."""
+    with PixlProducer(queue_name=TEST_QUEUE) as pp:
+        pp.publish(messages=[TEST_MESSAGE])
+        pp.publish(messages=[TEST_MESSAGE])
```
stefpiatek, line 43: probably can remove the second one that I added during debuggin

pixl_core/tests/patient_queue/test_subscriber.py
```diff
@@ -37,51 +33,29 @@
...
+        # Create a Task to run pc.run in the background
+        task = asyncio.create_task(pc.run(callback=consume))
+
+        # Wait for a short time to allow pc.run to start
+        await asyncio.sleep(1)
+
+        consume.assert_called()
+        task.cancel()
+        # need to close connection attribute or cancellation will hang waiting for next message
+        await pc._connection.close()  # noqa: SLF001
```
stefpiatek, line 56: oh that's annoying that we can't use the context handler to do it

pixl_core/tests/patient_queue/test_subscriber.py
```diff
@@ -37,51 +33,29 @@
...
+        await asyncio.sleep(1)  # Allow time for task to cancel
+    # pytest.fail("Check that async test runs")  # noqa: ERA001 uncomment for sanity check
```
stefpiatek, line 58: This might be a big rogue but I think is preferable to having uncommented code as we'd know if it ever changed
```suggestion
    pytest.xfail(reason="Sanity check that the test breaks out of async loop, so force failure")
        pytest.fail("Async has run, so exit with the only accepted failure")
```

stefpiatek (approved):
Thanks for cleaning up, some suggestions. Happy for you to merge with whatever you decide

pixl_core/tests/conftest.py
peshence: I had a (wrong) idea about fixing system tests.

pixl_core/tests/patient_queue/test_subscriber.py
peshence: I initially resolved this but then realised - yes, why isn't the handler, you know, handle it? So added connection close to __aexit__

peshence (commented):
Very good comments @stefpiatek I think significantly improved now

PR #283, opened 2024-02-06
#237 test[cli]: reduce docker reliance

peshence:
- remove tests that just wrap calls from core
- mock producer
- delete unnecessary docker compose and run tests script


cli/tests/docker-compose.yml
stefpiatek, line 1: so lovely

stefpiatek (approved):
👏 🚀

PR #285, opened 2024-02-07
Install module requirements before module in docker

stefpiatek:
Should allow better caching within docker layers.

Doesn't reduce reliance but makes docker builds faster for #237 

- Turns out you can just pip install from pyproject.toml without having any of the sourcecode
- Made python version in docker consistent and updated it to 3.11


milanmlft (approved):
🚀

PR #286, opened 2024-02-07
Update Rabbitmq version to 3.12

ruaridhg:
Updated rabbitmq version to 3.12-management in tests for modules:
- pixl imaging
- cli
- core
- system

Needed to `export DOCKER_DEFAULT_PLATFORM=linux/arm64` for it to work on Mac M1


pixl_imaging/tests/docker-compose.yml
```diff
@@ -54,6 +54,7 @@ services:
       - "9001:8000"
 
   vna-qr:
+    platform: linux/amd64
```
stefpiatek, line 57: are these all needed?

stefpiatek (approved):
Would be good to check that we need to specify the platform locally

pixl_imaging/tests/docker-compose.yml
ruaridhg: Having looked into it the answer appears to be yes. Also need to set `DOCKER_DEFAULT_PLATFORM=linux/arm64` for it to work locally which I'm not sure why.

Error when no platform is specified:
`Error response from daemon: image with reference osimis/orthanc:22.9.0-full-stable was found but does not match the specified platform: wanted linux/arm64, actual: linux/amd64 `

Error if I set `DOCKER_DEFAULT_PLATFORM=linux/amd64` (both with and without platform specified):
`Error response from daemon: image with reference rabbitmq:3.12.9-management was found but does not match the specified platform: wanted linux/amd64, actual: linux/arm64`

ruaridhg: Same pattern of errors for the system tests

stefpiatek: I think you'll need to down all of your containers and then do `docker system prune --all` as its downloaded that version on arm and doesn't seem to want to redownload it

milanmlft: Yeah, I don't think all that is necessary, I'm able to run these containers fine on my M2 (would surprise me if there would be a difference between M1-M2 in this case).
Make sure you have Rosetta enabled as well in Docker settings.
Screenshot 2024-02-09 at 09 52 17

stefpiatek:
Happy for you to emrge once you've resovled the conflicts

stefpiatek (approved):

PR #287, opened 2024-02-08
Delete test_producer and test both producer and subscriber togehter

ruaridhg:


stefpiatek:
going to close this one for now, can work out if we want to make this change later on

PR #288, opened 2024-02-08
Replace Docker FTPS server with `pytest` plugin fixture in `pixl_core` tests

milanmlft:
- Created an installable `pytest-pixl` plugin module to share fixtures across PIXL modules
- Added the `ftpserver` fixture to spin up a local FTPS server with `pyftpdlib`,  with support for implicit SSL
- Updated `pixl_core` tests to use the new fixture instead of the FTPS docker

Next steps will be to also integrate this in the system test and eventually migrate all shared `pytest` fixtures to the `pytest-pixl` module.


pixl_core/tests/conftest.py
```diff
@@ -68,6 +71,20 @@ def run_containers() -> subprocess.CompletedProcess[bytes]:
     )
 
 
+@pytest.fixture()
+def ftps_home_dir(ftps_server) -> Generator[Path, None, None]:
+    """Return the FTPS server home directory and clear it after every test"""
```
stefpiatek, line 76: does this use a tmpdir? If so then we don't need to clean up afterwards

pytest-pixl/src/pytest_pixl/ftpserver.py
```diff
@@ -0,0 +1,116 @@
+"""A ligthweight FTPS server supporting implicit SSL for use in PIXL tests."""
+
+from pathlib import Path
+
+from decouple import config
+from pyftpdlib.authorizers import DummyAuthorizer
+from pyftpdlib.handlers import TLS_FTPHandler
+from pyftpdlib.servers import FTPServer
+
+# USer permission
```
stefpiatek, line 10:
```suggestion
# User permission
```

.github/workflows/main.yml
```diff
@@ -85,7 +85,7 @@ jobs:
 
       - name: Install Python dependencies
         run: |
-          pip install pixl_core/[test]
+          pip install pytest-pixl pixl_core/[test]
```
stefpiatek, line 88: I wonder if this runs in order or dependencies, if so then it'll need to install core first

pytest-pixl/pyproject.toml
```diff
@@ -0,0 +1,40 @@
...
+  { name="PIXL core functionality" },
+]
+description = "Pytest plugin for PIXL"
+readme = "README.md"
+requires-python = ">=3.9"
+classifiers = [
+    "Programming Language :: Python :: 3"
+]
+dependencies = [
+    "core", # pixl_core
```
stefpiatek, line 14: can this just be a test dependency. Feels like it might get circular

stefpiatek (approved):
Looking good, not sure if we're going to get circular dependencies with core

PR #297, opened 2024-02-12
Test anonymised DICOMs can still be converted to NIFTI

jstutters:
Passes an MRI series through the PIXL DICOM anonymization then confirms that the image can still be converted to NIFTI and that the pre- and post- anonymization NIFTIs match.  Does _not_ test that an MRI DICOM can pass through the whole system.


testutils/README.md
```diff
@@ -0,0 +1,3 @@
+# testutils
```
stefpiatek, line 1: aha, might be worth looking at `pytest-pixl` and unifying with that. Having two separate test modules feels overkill

pixl_dcmd/tests/test_main.py
```diff
@@ -87,3 +90,49 @@ def test_pseudo_identifier_processing(rows_in_session, tag_scheme):
...
+        dcm_anon = apply_tag_scheme(dcm_identifiable, tag_scheme)
+        pydicom.dcmwrite(anon_dicom_dir / dcm_path.name, dcm_anon)
+
+    # Convert the anonymised DICOMs to NIFTI with dcm2niix
+    anon_nifti_dir = tmp_path / "nifti_anon"
+    anon_nifti_dir.mkdir()
+    subprocess.run(
+        ["dcm2niix", "-f", "anon", "-o", str(anon_nifti_dir), str(anon_dicom_dir)],
+        check=True,
+    )
```
stefpiatek, line 115: does the library offer a way in python to call it rather than using subprocess?

pixl_dcmd/tests/test_main.py
```diff
@@ -87,3 +90,49 @@ def test_pseudo_identifier_processing(rows_in_session, tag_scheme):
...
+        [
+            "dcm2niix",
+            "-f",
+            "ident",
+            "-o",
+            str(ident_nifti_dir),
+            str(directory_of_mri_dicoms),
+        ],
+        check=True,
+    )
```
stefpiatek, line 130: and again here

stefpiatek (changes requested):
Looking good, thanks Jon. I think we should work out a way of unifying this with `pytest-pixl` module, or at least moving both into a test-utils directory

testutils/README.md
jstutters: Yes, that makes a lot of sense.  At first glance testutils.dicom can be easily moved under pytest-pixl (which didn't exist when I started the branch in my defense!).

stefpiatek: Ha yep, no need for a defense there!

pixl_dcmd/tests/test_main.py
jstutters: See above.

pixl_dcmd/tests/test_main.py
stefpiatek: nah only if it was easy it would be nice, I think get it in asis

test/run-system-test.sh
```diff
@@ -28,8 +28,10 @@ docker compose --env-file .env -p system-test up --wait -d --build --remove-orph
 ./scripts/insert_test_data.sh
 
 # Install pixl_cli and test dependencies
-pip install -e "${PACKAGE_DIR}/pixl_core" && pip install -e "${PACKAGE_DIR}/cli"
-pip install pydicom
+pip install -e "${PACKAGE_DIR}/pixl_core" && \
+  pip install -e "${PACKAGE_DIR}/cli" && \
+  pip install -e "${PACKAGE_DIR}/testutils"
```
stefpiatek, line 33: I think this still needs to be updated

stefpiatek (approved):
May have reviewed too early but one dangling reference to testutils, then happy for you to merge

test/run-system-test.sh
jstutters: Certainly does! Fixed in commit 7cbaa2c9ef9551f0a2f757e907e9a70d5e822935.

HChughtai (approved):
Also looks good to me!

PR #298, opened 2024-02-12
Use alembic for database table creation

jstutters:
Allows us to be able to control the upgrade of the pixl database tables in production with upgrades and rollbacks


docker/postgres/Dockerfile
```diff
@@ -25,7 +25,8 @@ COPY --chmod=0755 ./postgres/postgres.conf /etc/postgresql/postgresql.conf
 
 COPY --chmod=0777 ./postgres/pixl-db_init.sh /docker-entrypoint-initdb.d/pixl-db_init.sh
 
-COPY --chmod=0777 ./postgres/create_pixl_tbls.py /pixl/create_pixl_tbls.py
+COPY alembic /pixl/alembic
+COPY alembic.ini /pixl/alembic.ini
```
stefpiatek, line 29: so I think instead of this being run on postgres spinning up, this should be part of the starting of the `imaging-api`, so we can remove installing pixl core here as well.

postgres/pixl-db_init.sh
```diff
@@ -25,4 +25,5 @@ psql -U "${POSTGRES_USER}" --dbname "${POSTGRES_DB}" -c "$ehr_create_command"
 
 source /pixl/venv/bin/activate
 
-python3 /pixl/create_pixl_tbls.py
+cd /pixl
+alembic upgrade head
```
stefpiatek, line 29: The otehr thing that we'd need to do for starting up the imaging API would be to have a check that 
if `SKIP_ALEMBIC=true`, then it doesn't run alembic (so that module level tests don't rely on postgres, but system tests do), and also set that ENV variable in the docker compose for the `imaging-api` tests

alembic/README
```diff
@@ -0,0 +1 @@
+Generic single-database configuration.
```
stefpiatek, line 1: Would be good to link to alembic docs and say that we're using it so that tables can change in production without losing of any data. Also worth updating the module level readme (for imaging-api, assuming we move it there)

stefpiatek (changes requested):
Really nice, thanks for your speedy work on this. I think we want to have this run when the `imaging-api` starts up because the postgres init db script only runs if there is no database. That way we can update the schema without losing any data

stefpiatek (approved):
Code changes look perfect, assuming it runs! I think there's also some documentation we could add to the alembic README. Happy for you to merge once you're happy with it

postgres/README.md
milanmlft, line 1: Maybe also record the `SKIP_ALEMBIC` env variable here? As in, why would one want to set this to `true`/`false`?

And not sure how familiar people generally are with `.mako` scripts, but maybe a brief overview how that works and/or some links to documentation?

milanmlft (commented):
Not really familiar with Alembic, so probably not the best judge but it looks good to me.
A bit more documentation would be nice though 😬

pixl_imaging/alembic/autogenerate-migration.sh
```diff
@@ -0,0 +1,44 @@
...
+fi
+
+
+# Set env variables
+export PIXL_DB_HOST=localhost
+export PIXL_DB_PORT=7654
+export POSTGRES_PORT=${PIXL_DB_PORT}
+export PIXL_DB_USER=pixl_db_username
+export PIXL_DB_PASSWORD=pixl_db_password
+export PIXL_DB_NAME=pixl
```
peshence, line 31: Why are these hardcoded? Shouldn't they come from some config file?

pixl_imaging/README.md
```diff
@@ -24,3 +24,21 @@ pip install -e ../pixl_core/ .
...
+
+## Configuration and database interaction
+
+The database tables are updated using alembic, see the [alembic](alembic) dir for more details.
+
+The `SKIP_ALEMBIC` environmental variable is used to control whether migrations are applied to the database.
+
+- Tests that don't use the database use `SKIP_ALEMBIC=true`, but otherwise you probably want to run this.
+- If you wanted to test out new migrations from a test/dev deployment on the GAE you may want to set
+skip alembic migrations
```
peshence, line 37: I don't get this line, it reads to me as "if you want to test migrations, skip them"

peshence (commented):
Looks good, huge improvement once it's over the line. Couple of small comments/questions here and there.

Also, shouldn't this be part of the core module?

pixl_imaging/alembic/autogenerate-migration.sh
stefpiatek: Yeah that's a fair point, for some reason i tried reading them in from the test .env file oddly they weren't getting picked up when alebmic ran. Let me see if I can resolve it

pixl_imaging/README.md
stefpiatek: That's fair, how does this look?
```suggestion
- If you want to ensure that a data migration setp works from the `pixl_test` or `pixl_dev` deployment on the GAE you have two choices:
1. pull in the new version with migrations and only redeploy the `pixl-imaging` container (keeping postgres up), or
2. Manually run each step yourself by skipping migrations being run on startup of the container 
  - pull in the code with the new migration(s)
```

peshence: I think I'm missing the reasoning for wanting to test the migration step and how that is different from normal deployment

stefpiatek: I guess if you have some changes to columns which require manpulating the data, you may want to test that if you had data in that table, that the migration is applied successfully. If that makes sense?

peshence: sort of, but why wouldn't this work with normal deployment? Do we wipe the database every time we deploy?

stefpiatek: it would. I guess the point is that only production would I expect to always be up, test and dev you'd probably wipe because you don't care about them. So this gives you a path to test this before rolling out to the production database. 

May be trying to prematurely optimise anyway, generally I expect we're not going to be doing migrations which require altering data. I'd expect we just add or remove columns and tables

milanmlft (approved):

PR #300, opened 2024-02-14
Update logging to be more useful

stefpiatek:
Adding in useful details for logging, decrease level of logging for areas that we don't really need to see during production, also some fixes for logging that never worked.

Orthanc doesn't show INFO by default (and doesn't have debug or trace AFAIK), so INFO is being used as DEBUG, WARN is being used as INFO, ERROR as WARN and ERROR 🤷 

Open to discussion if you think we'd need something as part of normal running. My rule of thumb is something like the [top answer here](https://stackoverflow.com/questions/186798/how-to-determine-what-log-level-to-use). 

One question is whether we want the healthcheck to be INFO. Given that the entire system is currently taken down if there's a failure this feels over the top, but I think it might be worth moving to allowing everything to run and have parts fail so at least we can finish processing the pipeline if there is an error in an earlier service (otherwise we may as well be running a monolith)


orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -237,7 +237,11 @@ def OnChange(changeType, level, resource):  # noqa: ARG001
         logger.info(msg)
         SendViaFTPS(resource)
 
-    if changeType == orthanc.ChangeType.ORTHANC_STARTED and _azure_available():
+    if (
+        changeType == orthanc.ChangeType.ORTHANC_STARTED
+        and _azure_available()
+        and ShouldAutoRoute()
```
milanmlft, line 243: Since we have an early return if `ShouldAutoRoute()` is false in https://github.com/UCLH-Foundry/PIXL/blob/9b17f44b473d9ad970e11af8704c72fb290a53b7/orthanc/orthanc-anon/plugin/pixl.py#L232, we don't actually need this check here I think? Same goes for line 235.

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -248,7 +252,7 @@ def OnChange(changeType, level, resource):  # noqa: ARG001
 
 def OnHeartBeat(output, uri, **request) -> Any:  # noqa: ARG001
     """Extends the REST API by registering a new route in the REST API"""
-    orthanc.LogWarning("OK")
+    orthanc.LogInfo("OK")
```
milanmlft, line 255: Agreed that this makes more sense as Info. No point in polluting the logs with a sequence of "OK"'s.

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -282,14 +287,14 @@ def AnonymiseCallback(dataset):
     tag operations through functions in pixl_dcmd module
     Returns writing anonymised dataset to disk
     """
-    orthanc.LogWarning("***Anonymising received instance***")
+    orthanc.LogWarning("Anonymising received instance")
```
milanmlft, line 290: Could make this `LogInfo` as well?

pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -53,7 +53,7 @@ async def process_message(message: Message) -> None:
     while job_state != "Success":
         if (time() - start_time) > config("PIXL_DICOM_TRANSFER_TIMEOUT", cast=float):
             msg = (
-                f"Failed to transfer {message.decode()} within "
+                f"Failed to transfer {message} within "
```
milanmlft, line 56: Nice, think this was overlooked after the `Message` refactoring 👍

milanmlft (approved):
🙌 Looking good! Just a small suggestion, which isn't actually related to the logging 🙂

orthanc/orthanc-anon/plugin/pixl.py
stefpiatek: ah yeah good call, if only I could read

orthanc/orthanc-anon/plugin/pixl.py
stefpiatek: Yeah its a fair question, felt like it'd be useful to know that we've started anonymisation in case we wanted to know how long the operation takes in production, or similar things. My rule of thumb is knowing when a substantial part you care about starts and ends. Not that strongly held though if you think otherwise?

PR #301, opened 2024-02-14
Refactor system tests to use the new FTP server and to run in pytest

### LINKED ISSUE #293: Replace flaky ftp server in system tests

- [ ] Use existing python-based ftp server in system tests
  - [x] Using a pytest fixture, so replace existing system tests to be run in pytest
  - [x] Configure pytest to display call captured stdout and stderr to the terminal live, while running the tests
  - [x] Use the `pixl_test` module to install dependencies required for testing rather than installing them in the shell script to run the system tests

### Testing

_No response_

### Documentation

- [ ] Update system tests documentation

### Dependencies

_No response_

### Details and Comments

Currently we have to rerun the system tests ~ half the time as the ftp server seems to exit on its own. Takes about 5 minutes to get to the point where it fails

jeremyestein:
To fix #293 

Well, this was a journey.
A simple refactor of the system (E2E) test ran into a few problems:
- the new, python-based FTP server hanging mysteriously
- container->host networking being awkward, but also differently so on linux and non-linux
- the FTP protocol itself (and python client implementation) interacting badly with the previous item (passive mode doesn't like NAT)

This explains some of the awkward workarounds in this PR.

Hopefully the benefits of moving it into pytest and having a non-flaky FTP server were worth it...

TODOs:
- [x] Refactor/dedupe "wait for X" type statements (including various TODO comments) -  Add some logging to say how long each bit waited for.
- [x] Be sure that we are waiting sufficiently for steps to finish before performing tests. Ie. no race conditions.
- [x] Improve DICOM image check to check the full file paths (it already *is* checking the full directory, if not the file name)
- [x] Turn on options to make TLS mandatory? (I can't find them any more and I don't think we need it anyway?)


pixl_core/src/core/upload.py
```diff
@@ -140,8 +141,8 @@ def _connect_to_ftp() -> FTP_TLS:
         ftp.login(ftp_user, ftp_password)
         ftp.prot_p()
     except ftplib.all_errors as ftp_error:
-        error_msg = "Failed to connect to FTPS server: '%s'"
-        raise ConnectionError(error_msg, ftp_error) from ftp_error
+        error_msg = f"Failed to connect to FTPS server: {ftp_user}@{ftp_host}:{ftp_port}"
```
milanmlft, line 144: Nice, this is definitely way better than my messy attempt to raise a connection error 😅

pytest-pixl/src/pytest_pixl/ftpserver.py
```diff
@@ -73,19 +73,20 @@ class PixlFTPServer:
...
+    def __init__(self, home_root: Path) -> None:
         """
         Initialise the FTPS server. Sets the hostname, port, username and password
         from the corresponding environment variables.
-        :param home_dir: The home directory for the FTP server. This is the directory where the
-            user will be placed after login.
+        :param home_root: The directory where the user's home directory will be created.
+        The home dir is the directory where the user will be placed after login.
         """
-        self.host = config("FTP_HOST", default="127.0.0.1")
```
milanmlft, line 83: Oh good call!

test/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+pytest_plugins = "pytest_pixl"
+
+
+@pytest.fixture()
+def host_export_root_dir():
+    """Intermediate export dir as seen from the host"""
+    return Path(__file__).parents[1] / "exports"
+
+
+def run_subprocess(
```
milanmlft, line 30: 😎

test/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+        ],
+        TEST_DIR,
+    )
+
+
+@pytest.fixture(scope="session")
+def _extract_radiology_reports(_setup_pixl_cli) -> None:
+    """
+    run pixl extract-radiology-reports.
+    TODO: should then poll for two minutes to ensure that the database is populated,
```
milanmlft, line 84: Could use a similar strategy as in https://github.com/UCLH-Foundry/PIXL/blob/545792b497975e4463f8afe970bae071b5f63884/test/scripts/check_max_storage_in_orthanc_raw.py#L43-L52? Maybe turn this in another helper in the `pytest` plugin?

test/run-system-test.sh
milanmlft, line 1: Love it!

milanmlft (approved):
🚀 This is looking great!

test/conftest.py
jeremyestein: Yes, this pattern crops up a few times, although unfortunately with slight variation so de-duplicating it will involve a bit of testing.

test/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+            check=True,
+            cwd=working_dir,
+            shell=shell,  # noqa: S603 input is trusted
+            timeout=timeout,
+            capture_output=True,
+        )
+    except subprocess.CalledProcessError as exception:
+        logging.error("*** exception occurred running: '%s'", cmd)  # noqa: TRY400 will raise anyway
+        logging.error("*** stdout:\n%s", exception.stdout.decode())  # noqa: TRY400
+        logging.error("*** stderr:\n%s", exception.stderr.decode())  # noqa: TRY400
```
peshence, line 50: Have you looked at whether logging.exception() would do the job here?

jeremyestein: The exception is being re-raised so will get printed eventually. The problem is that the stderr and stdout strings get shortened - only the beginning and end is shown - hence manually logging in their entirety here. I don't think I tried it but I'd expect `logging.exception` to have the same problem.

.github/workflows/main.yml
```diff
@@ -260,3 +266,14 @@ jobs:
...
           ./run-system-test.sh
+          echo FINISHED SYSTEM TEST SCRIPT
+
+      - name: Dump ehr-api docker logs for debugging
+        if: ${{ always() }}
+        run: |
+          docker logs -t system-test-ehr-api-1 2>&1
+
+      - name: Dump imaging-api docker logs for debugging
+        if: ${{ always() }}
```
stefpiatek, line 277: These are now failing because its running when the containers have been taken down
```suggestion
      - name: Dump ehr-api docker logs for debugging
        if: ${{ failure() }}
        run: |
          docker logs -t system-test-ehr-api-1 2>&1

      - name: Dump imaging-api docker logs for debugging
        if: ${{ failure() }}
```

test/check_utils.py
```diff
@@ -12,9 +11,5 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-set -euxo pipefail
 
-_sql_command="select * from emap_data.ehr_anon"
-_result=$(docker exec system-test-postgres-1 /bin/bash -c \
-  "PGPASSWORD=pixl_db_password psql -U pixl_db_username -d pixl -c \"$_sql_command\"")
-echo "$_result" | grep -q "2 row"
+"""Utilities for the system test"""
```
stefpiatek, line 15: should this have something in it?

test/conftest.py
stefpiatek: I think this function should live in an importable part of the package, so that we can re-use it in other places where we run subprocess. Hate to suggest `test/utils.py` but at least would be able to import it

test/run-system-test.sh
```diff
@@ -17,44 +17,56 @@ BIN_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
...
-./scripts/check_radiology_parquet.py \
-  ../exports/test-extract-uclh-omop-cdm/latest/radiology/radiology.parquet
-./scripts/check_ftps_upload.py
-
+setup () {
+  # The host OS determines how to reach the host from inside a container
+  if [ $(uname) = 'Linux' ]; then
+    export FTP_HOST=172.17.0.1
+  else
+    export FTP_HOST=host.docker.internal
```
stefpiatek, line 25: does using the container name work?
```suggestion
    export FTP_HOST=ftp-server
```

test/system_test.py
```diff
@@ -0,0 +1,135 @@
...
+            "/bin/bash",
+            "-c",
+            f'PGPASSWORD=pixl_db_password psql -U pixl_db_username -d pixl -c "{sql_command}"',
+        ],
+        Path.cwd(),
+    )
+    assert cp.stdout.decode().find("(2 rows)") != -1
+
+
+def test_max_storage_in_orthanc_raw():
```
stefpiatek, line 129: Imagine you're aware that this needs to be done, but just in case. I'm not super sure how useful these tests are, being able to create MRIs is useful so if its a pain with the order then don't spend long on it and drop them instead

stefpiatek (dismissed):
Looking good, not quite complete so requesting changes to avoid an overly keen merge

test/system_test.py
jeremyestein: Sorry I'm a bit lost, how is this connected to MRIs?

test/conftest.py
jeremyestein: Can do - could it go in `pytest_pixl`, or is that not really for that?

milanmlft: I think it could go in `pytest_pixl`, as long as it's not in `plugin.py` (as that is the entrypoint for the `pytest` plugin). Maybe something like `pytest_pixl/helpers.py`?

test/run-system-test.sh
stefpiatek: ha yes of course its not

test/system_test.py
stefpiatek: The tests create MRIs and send them into orthanc raw

test/conftest.py
stefpiatek: Yep that sounds good to me.

stefpiatek:
regarding the current system test error, I wonder if we should have part of the fixture is that we ping to see if the ftp server is up before yeilding. Its the first tests that are failing, which makes me think that the server process has started but its not connectable yet.

test/test_max_storage_in_orthanc_raw.py
```diff
@@ -0,0 +1,105 @@
...
+from decouple import config
+from pytest_pixl.dicom import write_volume
+
+SECONDS_WAIT = 5
+
+raw_instances_url = "http://localhost:{}/instances".format(config("ORTHANC_RAW_WEB_PORT"))
+
+
+@pytest.mark.usefixtures("_setup_pixl_cli")
+def test_max_storage_in_orthanc_raw():
```
jeremyestein, line 41: Can I confirm that it's this *entire file* (plus the `.sh` version in `test/scripts`) that we wanted to delete?

stefpiatek: yep!

test/check_utils.py
jeremyestein: It does now

test/conftest.py
jeremyestein: have moved it now

test/conftest.py
jeremyestein: I have made a generic waiting for condition method now.

jeremyestein:
Thanks to those who've already had a look at this. I believe it is now actually ready to go in.

docker-compose.yml
```diff
@@ -253,6 +256,10 @@ services:
       retries: 5
     networks:
       - pixl-net
+    # needed for testing under GHA (linux), so this container
+    # can reach the test FTP server running on the docker host
+    extra_hosts:
```
stefpiatek, line 261: If this is only needed for testing, I wonder if its worth making the build context for the test docker compose the same and then adding this only when running the test. If you think thats worth it, then create an issue but not blocking this PR

pytest-pixl/src/pytest_pixl/helpers.py
```diff
@@ -0,0 +1,92 @@
...
+from time import sleep
+from typing import TYPE_CHECKING, Callable, Optional
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+def run_subprocess(
```
stefpiatek, line 28: as we've got this, can we replace any other calls across tests with this function?

stefpiatek (approved):
Thanks for getting this in Jeremy, so much fun with FTP

milanmlft (approved):
🚀 works like a charm locally!

docker-compose.yml
jeremyestein: I did want to do this but didn't work out how.

pytest-pixl/src/pytest_pixl/helpers.py
jeremyestein: Have done all of them except one, which is using `text=True` which the wrapper doesn't support.

Issue #232, opened 2024-01-22
Configuration for multiple projects


stefpiatek:
### Definition of Done / Acceptance Criteria

PIXL pipeline should have configuration stored in files or database to be able to run the pipeline for multiple projects:

- [x] DICOM tag actions per project. Pointing to files (keeping 100 days and #284 being committed to the repo so we can use PRs)
- [x] whether images should be send via DICOMWeb or FTPS
- [x] FTPS connection details for project
- [ ] DICOMWeb connection details for sending data to endpoint

### Testing

No need to test DICOMWeb configuration at this stage, but will do this once we're sending via DICOMWeb API

### Documentation

Update documentation for how configuration gets set 

### Details and Comments

Currently we have single values set in the `.env` file or hardcoded details, but the PIXL pipeline should be able to be reused for multiple projects without any changes to the code

stefpiatek:
@timroberts would like to participate with this as there are touchpoints with OMOP ES, and @HChughtai with MIRSG & IoN DRI

milanmlft:
>to run the pipeline for multiple projects

Meaning we want to be able to do

```sh
pixl populate /resources/omop
pixl start
...
```

and have the whole pipeline run for multiple projects?

Or rather something like

```sh
pixl populate /resources/omop

# project 1
pixl start --config .project1_config

# project 2
pixl start --config .project2_config

# etc.
```

milanmlft:
As a side note, [Apple has released a new "configuration" programming language called Pkl]([Introducing Pkl, a programming language for configuration :: Pkl Docs](https://pkl-lang.org/blog/introducing-pkl.html)) which actually might be a good use case here. Though it's quite new, so not sure how stable it is 😬

stefpiatek:
> As a side note, [Apple has released a new "configuration" programming language called Pkl](https://pkl-lang.org/blog/introducing-pkl.html) which actually might be a good use case here. Though it's quite new, so not sure how stable it is 😬

That IS fun, but no python bindings. [dynaconf](https://www.dynaconf.com/) is a pretty nice way to deal with secrets in python

peshence:
Questions:
- The VOXL diagram implies parquet always moves to FTPS and images always move to DICOMWeb instances. Is that the plan?
- similar question - assuming FTPS is still an option for images, is the FTPS setting the same for parquet and for images? IE they are different per project
[...]

PR #302, opened 2024-02-14
Add multi-project config design diagram

milanmlft:
Adds a diagram with the current design proposal for supporting mult-project configuration.
Related to #232.


peshence (approved):
Looks good to me

docs/design/diagrams/pixl-multi-project-design.png
stefpiatek, line 1: The diagram looks good, we should add to the issue to write up design notes for this and what we've considered. Some changes but happy for you to merge once you've resolved them

Some notes from yesterday's chat which I think affect the diagram:


- config file
  - [ ] I don't think parquet files would be sent via dicomweb. Azure blob storage is a good shout though.
- Azure key vault
  - [ ] Worth representing the hierarchical nature, common practive is double underscore so `my_project__FTP__HOST`	
  - [ ] Not just could, but default expectation is that a projects endpoint is unique as it'll probably have its own auth. We could have aliases if we reused a connection e.g. `xnat__DICOMWEB__HOST` etc. if we had a system which could take multiple projects and manage the authorisation  
- [ ] I'd also be tempted to have the uploader API bind the PIXL projects dir, so it reads from there rather than having to upload all details into the database (just needs the project name)

stefpiatek (approved):
The diagram looks good, we should add to the issue to write up design notes for this and what we've considered. Some changes but happy for you to merge once you've resolved them

PR #304, opened 2024-02-15
Experimenting with Pkl for the config template

milanmlft:
Learned how to `Pkl` config files and gave it a try to support our config template.
Still very rough at this point obviously, maybe `Pkl` is a bit overkill.
I'd say the main advantage is the straightforward Type Annotation, and restricting the permitted values for a given field. 

Happy to look for other solutions though if people are not convinced 🙂 

Related to #232

### DIFF
--- a/config-template/PixlConfig.pkl
+++ b/config-template/PixlConfig.pkl
@@ -1,0 +1,22 @@
+module PixlConfig
+
+project: String
+
+class DicomAnonProfile {
+    base_profile: String // TODO: should this be a file path?
+    extension_profile: String? // optional
+}
+
+dicom_anon_profile: DicomAnonProfile
+
+// Permitted values for the destination types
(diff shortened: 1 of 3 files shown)


milanmlft:
Oh and not sure yet what should go into the `dicom_anon_profile`, so any feedback on that more than welcome as well!

stefpiatek:
Happy to have a chat about this, but feels like we're using our "novelty chips" on something which isn't that complex? [toml-schema](https://github.com/brunoborges/toml-schema) exists and means people don't have to learn another markup language for a pretty straightforward file.

With the novelty chips, the basic idea is that you should choose the standard option most of the time and then maybe 3 times in a project push the boundaries of what you use. So that you still do some fun new things, but you limit them

milanmlft:
Closing as we'll not move forward with Pkl.

PR #309, opened 2024-02-19
Support configuration for multiple projects

### LINKED ISSUE #232: Configuration for multiple projects

milanmlft:
## Current status

- Created an example project config file in `config-template/example-config.yml`
- Implemented `pydantic` model of project configuration
- Added some tests to test the `pydantic` model
- Refactored `core.upload` so it now has an `Uploader` interface for uploading DICOM and Parquet files to the right destination depending on the project's config
- Fetch project secrets, such as credentials for the uploader, from Azure Key Vault

## To do

- [x] Update system tests, but probably makes sense to wait for #301 to be merged
- [x] Add `.secrets.env` with the relevant envvars on GitHub CI

***
Fixes #232


peshence:
Looks great! I've not used pydantic before, pretty cool!

I would consider having the tag operations be an array to allow for multilevel configuration sharing, as we could have e.g.
- base profile for all modalities
- extension profile for X-Rays
- extension profile specific to this project

or have groups of tags, or something else. Just an idea though, not sure if it actually reflects the project needs, this should be enough for the time being

.env.sample
```diff
@@ -97,7 +93,5 @@ RABBITMQ_PASSWORD=
 PIXL_DICOM_TRANSFER_TIMEOUT=240
 PIXL_QUERY_TIMEOUT=10
 
-# FTP server
-FTP_HOST=
-FTP_USER_NAME=
-FTP_USER_PASSWORD=
+# Project configs directory
+PROJECT_CONFIGS_DIR=
```
stefpiatek, line 97: we should set this with a default value, assuming it can take relative paths

README.md
```diff
@@ -188,6 +188,141 @@ For the deidentification of the EHR extracts, we rely on an instance running the
...
+
+- Project name: the `<project-slug>` name of the Project
+- The DICOM dataset modalities to retain (e.g. `["DX", "CR"]` for X-Ray studies)
+- The anonymisation operations to be applied to the DICOM tags, by providing a file path to one or multiple YAML files
+- The endpoints used to upload the anonymised DICOM data and the public and radiology
+  [parquet files](./docs/file_types/parquet_files.md). We currently support the following endpoints:
+    - `"none"`: no upload
+    - `"ftps"`: a secure FTP server (for both _DICOM_ and _parquet_ files)
+      Requires the `FTPS_*` environment variables to be set in `.env`
+    - `"azure"`: a secure Azure Blob Storage account (for both _DICOM_ and _parquet_ files)
```
stefpiatek, line 224: for azure we wouldn't store dicom in blob storage, we'd use the dicom web service. Though I think for now we should only have options which we implement, so keep let's keep to FTPS and then for VOXL we'll bring in dicomweb?

README.md
```diff
@@ -188,6 +188,141 @@ For the deidentification of the EHR extracts, we rely on an instance running the
...
+The configuration file defines:
+
+- Project name: the `<project-slug>` name of the Project
+- The DICOM dataset modalities to retain (e.g. `["DX", "CR"]` for X-Ray studies)
+- The anonymisation operations to be applied to the DICOM tags, by providing a file path to one or multiple YAML files
+- The endpoints used to upload the anonymised DICOM data and the public and radiology
+  [parquet files](./docs/file_types/parquet_files.md). We currently support the following endpoints:
+    - `"none"`: no upload
+    - `"ftps"`: a secure FTP server (for both _DICOM_ and _parquet_ files)
+      Requires the `FTPS_*` environment variables to be set in `.env`
```
stefpiatek, line 223: none of these should require any `.env` variables though right?

README.md
```diff
@@ -188,6 +188,141 @@ For the deidentification of the EHR extracts, we rely on an instance running the
...
+```bash
+az keyvault create --resource-group <resource-group-name> --name <key-vault-name> --location "UKSouth"
+```
+
+### Step 2
+
+Create Service Principal & grant access as per
+
+```bash
+az ad sp create-for-rbac -n hasher-api --skip-assignment
```
stefpiatek, line 274: Not `hasher-api`, `pixl-secrets` is the name we used for this

README.md
```diff
@@ -188,6 +188,141 @@ For the deidentification of the EHR extracts, we rely on an instance running the
...
+
+copy the secret and paste as <secret-value> below
+
+```bash
+az keyvault secret set --vault-name "<key-vault-name>" --name "<secret-name>" --value "<secret-value>"
+```
+
+### Step 5
+
+Save credentials in `.secrets.env` and a LastPass `PIXL Keyvault <environment> secrets` note.
```
stefpiatek, line 316: have we done this step?

cli/README.md
```diff
@@ -141,9 +141,10 @@ pip install -e ../pixl_core/ -e .[test]
 
 ### Running tests
 
-Tests can be run with `pytest` from the `tests` directory.
+The CLI tests require a running instance of the `rabbitmq` service, for which we provide a
+`docker-compose` [file](./tests/docker-compose.yml). The service is automatically started by the
+`run_containers` _pytest_ fixture. So to run the tests, simply run
```
stefpiatek, line 146: A bit carpenties of a suggestion but
```suggestion
`run_containers` _pytest_ fixture. So to run the tests, run
```

docker-compose.yml
```diff
@@ -20,311 +20,342 @@ x-http-proxy: &http-proxy ${HTTP_PROXY}
 x-https-proxy: &https-proxy ${HTTPS_PROXY}
 x-no-proxy: &no-proxy localhost,0.0.0.0,127.0.0.1,uclvlddpragae07,hasher-api,orthanc-raw
 x-proxy-common: &proxy-common
-  HTTP_PROXY: *http-proxy
-  http_proxy: *http-proxy
-  HTTPS_PROXY: *https-proxy
-  https_proxy: *https-proxy
-  NO_PROXY: *no-proxy
-  no_proxy: *no-proxy
+    HTTP_PROXY: *http-proxy
```
stefpiatek, line 23: these are a lot of tabs that I'm guessing weren't intentional. Don't have strong opinions on spacing but we probably don't want to be mixing tabs and spaces

docs/design/diagrams/pixl-multi-project-config.png
stefpiatek, line 1: the key naming is different because I forgot that you can't have underscores in key names

pixl_core/src/core/exports.py
```diff
@@ -110,3 +113,17 @@ def export_radiology(self, export_df: pd.DataFrame) -> pathlib.Path:
...
+        """Upload the latest extract to the DSH."""
+        project_config = load_project_config(self.project_slug)
+        destination = project_config.destination.parquet
+
+        if destination == "ftps":
+            ftps_uploader = FTPSUploader(project_config)
+            msg = f"Uploading parquet files for project {self.project_slug} via FTPS"
+            logger.warning(msg)
+            ftps_uploader.upload_parquet_files(self)
+        else:
```
stefpiatek, line 127: I think if None then we should just log as a warning that its not set to export

pixl_core/src/core/upload.py
```diff
@@ -11,179 +11,149 @@
...
+        """
+        Initialise the uploader for a specific project with the destination configuration and an
+        AzureKeyvault instance. The keyvault is used to fetch the secrets required to connect to
+        the remote destination.
+
+        :param project: The project name for which the uploader is being initialised. Used to fetch
+            the correct secrets from the keyvault.
+        """
+        self.project_config = project_config
+        self.keyvault = AzureKeyVault()
```
stefpiatek, line 57: If we're going to make an abstract base class then why don't we have methods like _set_config which you have to override, and upload_dicom etc. Like we'll always want to run _set_config on initialisation right?

You could even have a static method in this class which takes in the project config and creates the right child class and returns it. That way there's a single point for parsing the config value for both dicom and parquet?

projects/configs/uclh-nasogastric-tube-project-ngt-only-full-dataset.yaml
```diff
@@ -0,0 +1,24 @@
...
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+project:
+    name: "uclh-nasogastric-tube-project-ngt-only-full-dataset"
+    modalities: ["DX", "CR"]
+
+tag_operation_files:
+    ["uclh-nasogastric-tube-project-ngt-only-full-dataset-tag-operations.yaml"]
```
stefpiatek, line 20: is there a benefit to having this ending with `tag-operations`?

template_config.yaml
```diff
@@ -0,0 +1,23 @@
...
+#  limitations under the License.
+
+project:
+    name: "project-slug"
+    modalities: ["DX", "CR"] # DICOM dataset modalities to retain
+
+tag_operation_files: ["base-tag-operations.yaml"] # DICOM tag anonymisation operations, can specify multiple files
+
+destination:
+    dicom: "ftps" # alternatives: "none", "dicomweb", "azure"
```
stefpiatek, line 22: Just adding here as well, I don't think we should be offering options which aren't implemented

test/run-system-test.sh
```diff
@@ -17,36 +17,35 @@ BIN_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
...
-  # Warning: Requires to be run from the project root
-  (cd "${PACKAGE_DIR}" && \
-    docker compose --env-file test/.env -p system-test up --wait -d --build)
+setup() {
+	docker compose --env-file .env -p system-test down --volumes
+	#
+	# Note: cannot run as single docker compose command due to different build contexts
+	docker compose --env-file .env -p system-test up --wait -d --build --remove-orphans
+	# Warning: Requires to be run from the project root
+	(cd "${PACKAGE_DIR}" &&
```
stefpiatek, line 26: I guess because its in brackets it doesn't need the line continuation?

stefpiatek (approved):
Looks good, fair few bits that I think we can clean up, happy for you to merge after you're happy with them. In hindsight, probably should have broken this up into smaller PRs

stefpiatek:
>     * The `test--ftp--host` secret in the Azure Keyvault is set to `"ftp-server"`, but this clashes with the system test which assumes the FTP server is `"localhost"`
>       → [this is what is currently making the system test fail](https://github.com/UCLH-Foundry/PIXL/actions/runs/8055569203/job/22002718134?pr=309#step:13:33)

Updated the port and host

>       * (less severe) the `test--ftp--*` secret names don't match the project slug of our test project (`"test-extract-uclh-omop-cdm"`). I worked around this by allowing to set an alias in the project config file, but we might consider updating the Keyvault secret names

Why don't we rename the project to `test`? As we're only using it for testing?

 
>     * [Configuration for multiple projects #232 (comment)](https://github.com/UCLH-Foundry/PIXL/issues/232#issuecomment-1964002282)_ EXPORT_AZ_NAME

I'm not sure, why did you add it?


> Also slightly worrying are the [several warnings](https://github.com/UCLH-Foundry/PIXL/actions/runs/8055569203/job/22002718134?pr=309#step:14:54) in the `orthanc-anon` logs during the system test about missing tag operations (?). Not sure where that is coming from.

Not worried about that, we've got dumb dummy data that doesn't have all of the tags that we're annonymising

pixl_core/src/core/upload.py
```diff
@@ -11,178 +11,144 @@
...
-    def sock(self, value: socket) -> None:
-        """When modifying the socket, ensure that it is ssl wrapped."""
-        if value is not None and not isinstance(value, ssl.SSLSocket):
-            value = self.context.wrap_socket(value)
-        self._sock = value
+    def _set_config(self) -> None:
+        self.host = self.keyvault.fetch_secret(f"{self.project}--ftp--host")
+        self.user = self.keyvault.fetch_secret(f"{self.project}--ftp--username")
+        self.password = self.keyvault.fetch_secret(f"{self.project}--ftp--password")
+        self.port = config("FTP_PORT", default=21, cast=int)
```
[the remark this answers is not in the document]
milanmlft: Do we have it as a secret? Thought we were only storing host, username and password

README.md
milanmlft: You mean the LastPass? Guess it wouldn't be a bad idea to track projects? But maybe change "`<environment>`" to `<project>`

README.md
```diff
@@ -188,6 +188,141 @@ For the deidentification of the EHR extracts, we rely on an instance running the
...
+    git pull
+    git switch -c <branch-name>
+    ```
+
+1. Copy the `template_config.yaml` file to a new file in the `projects/config` directory and fill
+   in the details.
+1. The filename of the project config should be `<project-slug>`.yaml
+
+    >[!NOTE]
+    > The project slug should match the project name in the `extract_summary.json` log file!
```
peshence, line 208: Perhaps suggest using slugify to get a consistent result?

docker-compose.yml
milanmlft: Ah crap, that's my autoformatter (though I do like 4-space indents)
But I don't see any tabs, just 4-space indents?

README.md
```diff
@@ -188,6 +188,141 @@ For the deidentification of the EHR extracts, we rely on an instance running the
...
+
+## Azure Keyvault setup
+
+_This is done for the \_UCLH_DIF_ `dev` tenancy, will need to be done once in the _UCLHAZ_ `prod`
+tenancy when ready to deploy to production.\_
+
+This Key Vault and secret must persist any infrastructure changes so should be separate from disposable
+infrastructure services. ServicePrincipal is required to connect to the Key Vault.
+
+The application uses the ServicePrincipal and password to authenticates with Azure via environment
```
peshence, line 258:
```suggestion
The application uses the ServicePrincipal and password to authenticate with Azure via environment
```
Typo. Also, ServicePrincipal is used both as a name and as a regular noun within a couple of sentences which makes it confusing. Maybe a short explanation of what it is might help?

cli/tests/conftest.py
```diff
@@ -22,6 +22,8 @@
 from sqlalchemy import Engine, create_engine
 from sqlalchemy.orm import Session, sessionmaker
 
+os.environ["PROJECT_CONFIGS_DIR"] = str(pathlib.Path(__file__).parents[2] / "projects/configs")
```
peshence, line 25: I know I wrote this, but what about making the configs folder a python module and using pip and importlib to access it instead of this shaky parents[2] structure? Then we don't need to install editable either

docker-compose.yml
peshence: I think you've just switched it from 2 to 4 spaces. Mine does it too, and I'd argue it's the standard/cleaner, but we should keep it the same across all our yml files

pixl_core/src/core/upload.py
milanmlft: Yeah I didn't want to put too much in the base class because I wasn't sure how similar/different the other upload strategies would be for Azure and DICOMweb as we don't really have a plan for those yet AFAIK?
But the `_set_config` makes sense. And maybe `upload_dicom` as well, but `upload_parquet_files` would only exist for some upload strategies.

pixl_core/src/core/project_config.py
```diff
@@ -0,0 +1,103 @@
...
+    ftps = "ftps"
+    azure = "azure"
+    dicomweb = "dicomweb"
+
+
+class _Destination(BaseModel):
+    dicom: _DestinationEnum
+    parquet: _DestinationEnum
+
+    @validator("parquet")
```
peshence, line 75: `validator` is deprecated as of pydantic v2 (supported on python >=3.8), so I would prefer to not add it and replace with `field_validator`. We had a separate discussion with @milanmlft as he gets a clash with some other library that I don't. I also couldn't find where we set the pydantic depdndency (and therefore version), perhaps we can do some pair programming to figure this one out.

pixl_core/src/core/upload.py
```diff
@@ -11,179 +11,149 @@
...
     from core.exports import ParquetExport
+    from core.project_config import PixlConfig
 
 
-from core.db.queries import get_project_slug_from_db, update_exported_at
+from core._secrets import AzureKeyVault
+from core._upload_ftps import (
+    _connect_to_ftp,
+    _create_and_set_as_cwd,
+    _create_and_set_as_cwd_multi_path,
```
peshence, line 36: really odd to be importing private functions, maybe we should rethink this a bit?

pixl_core/src/core/upload.py
stefpiatek: yeah can always have upload partquet files return raise a not implemented error for those types?

pixl_core/tests/test_upload.py
peshence, line 1: Worth renaming to test_ftps_upload?

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -76,7 +138,7 @@ def remove_overlays(dataset: Dataset) -> Dataset:
     return dataset
 
 
-def enforce_whitelist(dataset: dict, tags: dict) -> dict:
+def enforce_whitelist(dataset: dict, tags: list[dict]) -> dict:
```
peshence, line 141: I think this and the next return list[dict] as well?

projects/configs/uclh-nasogastric-tube-project-ngt-only-full-dataset.yaml
peshence: not really I guess, I always prefer not to have duplicate file names, but I'm happy to change if you think this is better

template_config.yaml
```diff
@@ -0,0 +1,23 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+project:
+    name: "project-slug"
+    modalities: ["DX", "CR"] # DICOM dataset modalities to retain
+
+tag_operation_files: ["base-tag-operations.yaml"] # DICOM tag anonymisation operations, can specify multiple files
```
peshence, line 19: currently we actually cannot
```suggestion
tag_operation_files: ["base-tag-operations.yaml"] # DICOM tag anonymisation operations, can specify multiple files in the future 
```

test/run-system-test.sh
peshence, line 1: formatting got messed up here, it's some form of a tab now as far as I can tell but at least on github visually it looks like >=8 spaces

peshence (commented):
Looks good, some comments for @milanmlft , some for myself

test/run-system-test.sh
milanmlft: 🤦‍♂️ guess that's what I get for overconfiguring my neovim setup

pixl_core/src/core/upload.py
peshence: I'd argue for 
- moving the FTPSUploader child class into the _upload_ftps module 
- making the _upload_ftps module public
- making ImplicitFtpTls a private class

cli/tests/conftest.py
milanmlft: Sounds good, but maybe make a separate issue and PR for it? Because this PR is getting quite bloated already as it is.

pixl_core/src/core/_secrets.py
```diff
@@ -0,0 +1,79 @@
...
+        self.kv_name = config("AZURE_KEY_VAULT_NAME")
+        self.client = self._connect_to_keyvault()
+
+    def _connect_to_keyvault(self) -> SecretClient:
+        key_vault_uri = f"https://{self.kv_name}.vault.azure.net"
+
+        credentials = DefaultAzureCredential()
+        return SecretClient(vault_url=key_vault_uri, credential=credentials)
+
+    def fetch_secret(self, secret_name: str) -> str:
```
stefpiatek, line 51: We should decorate this with `@lru_cache` so we won't hit the keyvault in a long running session

.github/workflows/main.yml
```diff
@@ -223,6 +223,18 @@ jobs:
...
+        env:
+          EXPORT_AZ_CLIENT_ID: ${{ secrets.EXPORT_AZ_CLIENT_ID }}
+          EXPORT_AZ_CLIENT_PASSWORD: ${{ secrets.EXPORT_AZ_CLIENT_PASSWORD }}
+          EXPORT_AZ_TENANT_ID: ${{ secrets.EXPORT_AZ_TENANT_ID }}
+          EXPORT_AZ_KEY_VAULT_NAME: ${{ secrets.EXPORT_AZ_KEY_VAULT_NAME }}
+        run: |
+          echo EXPORT_AZ_CLIENT_ID=$EXPORT_AZ_CLIENT_ID >> .secrets.env
+          echo EXPORT_AZ_CLIENT_PASSWORD=$EXPORT_AZ_CLIENT_PASSWORD >> .secrets.env
+          echo EXPORT_AZ_TENANT_ID=$EXPORT_AZ_TENANT_ID >> .secrets.env
+          echo EXPORT_AZ_KEY_VAULT_NAME=$EXPORT_AZ_KEY_VAULT_NAME >> .secrets.env
```
stefpiatek, line 236: tbf I _think_ you could just create an empty .secrets.env and then make sure that these variables are in use when we run the tests

pixl_core/src/core/upload.py
milanmlft: Agreed!

pixl_core/tests/test_upload.py
milanmlft: I like having some consistency between the `test_<module_name>` files and the modules they're testing. But if we move around the `FTPSUploader` and put in its own `ftps_uploader.py` module, I think it makes sense to rename this test file as well 🙂

cli/tests/conftest.py
peshence: created #326

pixl_dcmd/src/pixl_dcmd/main.py
peshence: it is in fact a Dataset as both input and output

cli/tests/test_check_env.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_check_env_fails(tmp_path):
+    """
+    Test that check_env fails when the current test env file does not match the sample env file.
+    """  # noqa: D200 either this or it's 102 chars
+    tmp_sample_env_file = tmp_path / ".env.sample"
+    tmp_sample_env_file.write_text("NONEXISTENT_VARIABLE=")
+
+    runner = CliRunner()
+    result = runner.invoke(check_env, tmp_sample_env_file.as_posix())
```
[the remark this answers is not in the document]
peshence: on my list

docker-compose.yml
peshence, line 1: this commit did more than it says, I've noticed missing variables and wrong mounts, might be other stuff too

.github/workflows/main.yml
```diff
@@ -182,6 +182,11 @@ jobs:
     if: ${{ ! github.event.pull_request.draft || contains(github.event.pull_request.title, '[force-system-test]') }}
     runs-on: ubuntu-22.04
     timeout-minutes: 30
+    env:
+      EXPORT_AZ_CLIENT_ID: ${{ secrets.EXPORT_AZ_CLIENT_ID }}
+      EXPORT_AZ_CLIENT_PASSWORD: ${{ secrets.EXPORT_AZ_CLIENT_PASSWORD }}
+      EXPORT_AZ_TENANT_ID: ${{ secrets.EXPORT_AZ_TENANT_ID }}
+      EXPORT_AZ_KEY_VAULT_NAME: ${{ secrets.EXPORT_AZ_KEY_VAULT_NAME }}
```
stefpiatek, line 189: We should restrict the scope these are available I reckon. Does it work if we only set them for the steps where we run the tests?

README.md
```diff
@@ -188,6 +188,148 @@ For the deidentification of the EHR extracts, we rely on an instance running the
...
+
+Save credentials in `.secrets.env` and a LastPass `PIXL Keyvault <project-slug> secrets` note.
+
+```
+EXPORT_AZ_CLIENT_ID=<generated-app-ID>
+EXPORT_AZ_CLIENT_PASSWORD=<generated-password>
+EXPORT_AZ_TENANT_ID=<tenant-ID>
+EXPORT_AZ_KEY_VAULT_NAME=<key-vault-name>
+```
+</details>
```
stefpiatek, line 331: At a quick search I don't think we've done this step, can you add it to lastpass?

docker-compose.yml
```diff
@@ -55,11 +55,11 @@ x-pixl-db: &pixl-db
   PIXL_DB_PASSWORD: ${PIXL_DB_PASSWORD}
   PIXL_DB_NAME: ${PIXL_DB_NAME}
 
-x-ftp-host: &ftp-host
-  FTP_HOST: ${FTP_HOST}
-  FTP_USER_NAME: ${FTP_USER_NAME}
-  FTP_USER_PASSWORD: ${FTP_USER_PASSWORD}
-  FTP_PORT: ${FTP_PORT}
+x-azure-keyvault: &azure-keyvault
```
stefpiatek, line 58: love to see it

stefpiatek (approved):
Really nice, thanks both 🎉

README.md
peshence: Yes we have, it's called pixl_screts.env

pixl_core/src/core/project_config.py
```diff
@@ -0,0 +1,103 @@
...
+    azure_kv_alias: Optional[str]
+    modalities: list[str]
+
+
+class _DestinationEnum(str, Enum):
+    """Defines the valid upload destinations."""
+
+    none = "none"
+    ftps = "ftps"
+    azure = "azure"
```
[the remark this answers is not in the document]
stefpiatek: I think sftp and none for both

README.md
stefpiatek: okay, can we change the documentation do reflect this, because its not to do with the <project-slug>. Maybe call it `pixl_dev_secrets.env`?

.github/workflows/main.yml
peshence: I don't see how that'd work if we need these env variables to be in several different docker containers

milanmlft: Ah yes, good point

README.md
milanmlft: Ah yes I updated the original `pixl-secrets` note with this, as it was causing some confusion.
But I can create a proper project note as well.

PR #311, opened 2024-02-20
Get rid of `pixl_config.yml`

milanmlft:
Attempt to reduce complexity by:

- Removing `pixl_config.yml` and moving the useful bits to `.env` (which already largely had the information contained in `pixl_config.yml`)
- Move the `APIConfig` class to `cli._config`
- Reduce the use of the name `config`, to avoid confusion and in preparation of #309 

**Motivation**: in #309 we will add `config.yml` files to configure how projects should be run through the pipeline. The information contained in those `config.yml` files are completely different from what is in the current `pixl_config.yml` file, which may lead to some confusion. In addition, the information in `pixl_config.yml` is already encoded in the `.env` environment variables anyway.

***

## To do

- [x] Check that system test still succeeds
- [x] Remove all references to `pixl_config.yml` in the docs
- [x] Document the appropriate way of configuring the EHR and Imaging APIs with the new env variables


cli/src/pixl_cli/_config.py
```diff
@@ -13,20 +13,57 @@
...
-import yaml
+from decouple import config
 
+API_SETTINGS = {"ehr_api": {}, "imaging_api": {}}  # type: dict
+API_SETTINGS["ehr_api"]["host"] = config("PIXL_EHR_API_HOST", default="localhost")
+API_SETTINGS["ehr_api"]["port"] = int(config("PIXL_EHR_API_PORT", default=7006))
+API_SETTINGS["ehr_api"]["default_rate"] = int(config("PIXL_EHR_API_RATE", default=1))
+API_SETTINGS["imaging_api"]["host"] = config("PIXL_IMAGING_API_HOST", default="localhost")
+API_SETTINGS["imaging_api"]["port"] = int(config("PIXL_IMAGING_API_PORT", default=7007))
+API_SETTINGS["imaging_api"]["default_rate"] = int(config("PIXL_IMAGING_API_RATE", default=1))
```
stefpiatek, line 26: If we're going to have this as a constant then we should initialise it all in one, so like this? May be a moot point if we create the API settings objects directly
```suggestion
API_SETTINGS = {
  "ehr_api": {
    "host": config("PIXL_EHR_API_HOST", default="localhost"),
    "port": int(config("PIXL_EHR_API_PORT", default=7006))
    "default_rate": int(config("PIXL_EHR_API_RATE", default=1))
    },
  }, 
  "imaging_api": {...}
  }

API_SETTINGS["imaging_api"]["host"] = config("PIXL_IMAGING_API_HOST", default="localhost")
API_SETTINGS["imaging_api"]["port"] = int(config("PIXL_IMAGING_API_PORT", default=7007))
API_SETTINGS["imaging_api"]["default_rate"] = int(config("PIXL_IMAGING_API_RATE", default=1))
```

cli/src/pixl_cli/_config.py
```diff
@@ -13,20 +13,57 @@
...
+SERVICE_SETTINGS["rabbitmq"]["password"] = config("RABBITMQ_PASSWORD", default="rabbitmq_password")
 
-    with Path(filename).open() as config_file:
-        config_dict = yaml.safe_load(config_file)
-    return dict(config_dict)
+SERVICE_SETTINGS["postgres"]["host"] = config("POSTGRES_HOST", default="localhost")
+SERVICE_SETTINGS["postgres"]["port"] = int(config("POSTGRES_PORT", default=7001))
+SERVICE_SETTINGS["postgres"]["username"] = config("PIXL_DB_USER", default="pixl_db_username")
+SERVICE_SETTINGS["postgres"]["password"] = config("PIXL_DB_PASSWORD", default="pixl_db_password")
+SERVICE_SETTINGS["postgres"]["database"] = config("PIXL_DB_NAME", default="pixl")
```
stefpiatek, line 39: Is it worth building all of the API settings objects here rather than importing the env dicts?

cli/src/pixl_cli/_config.py
```diff
@@ -13,20 +13,57 @@
...
+
+        self.__dict__.update(kwargs)
+
+    @property
+    def base_url(self) -> str:
+        """Return the base url for the API"""
+        return f"http://{self.host}:{self.port}"
+
+
+def api_config_for_queue(queue_name: str) -> APIConfig:
```
stefpiatek, line 58: if we did pre-build the objects then they'd be in a dictionary like this
```python
_api_configs = {"imaging": imagingApiConfig, "ehr": ehrApiConfig} and we'd check to see if the queue name exists in the dictionary and if not raise the error, otherwise return the object
```

cli/src/pixl_cli/_config.py
```diff
@@ -13,20 +13,66 @@
 #  limitations under the License.
 
 """Configuration of CLI from config file."""
-from pathlib import Path
+from typing import Optional
 
-import yaml
+from decouple import config
 
+SERVICE_SETTINGS = {
```
stefpiatek, line 20: One thing here, I don't think any of these should have defaults that aren't set in the `.env.sample`, because then if I forget to set one it'll look like I have the config set up

stefpiatek (approved):
Works on GAE. What's more fun is that if you're using the environment then it can be done from any directory if you've used pip install -e. Which I think is okay, or do we want it to read the config only from the directory that the cli is initiated in?

stefpiatek:
>  I would say it's fine, not much chance that there would be any name clashes for these environment variables.

Only risk is if you use the environment for the wrong project. e.g. I'm in dev but I use the test environment. Agree that I think its fine

cli/src/pixl_cli/_config.py
milanmlft: Makes sense, the only annoying thing is that you will then have to source the `.env` file first every time before running any of the `pixl` CLI commands 😕

stefpiatek: Why would you have to do that? Doesn't the PIXL CLI read in the .env file as part of the config?

milanmlft: Don't think it does? At least not when I run it from the `test/` directory in the system-test:

```sh
+ pixl populate /Users/milan/Projects/UCLH-Foundry/PIXL/test/resources/omop
Traceback (most recent call last):
  File "/Users/milan/.pyenv/versions/3.10.13/envs/pixl-3.10.13/bin/pixl", line 5, in <module>
    from pixl_cli.main import cli
  File "/Users/milan/Projects/UCLH-Foundry/PIXL/cli/src/pixl_cli/__init__.py", line 17, in <module>
    from pixl_cli.main import cli
  File "/Users/milan/Projects/UCLH-Foundry/PIXL/cli/src/pixl_cli/main.py", line 28, in <module>
    from pixl_cli._config import SERVICE_SETTINGS, api_config_for_queue
  File "/Users/milan/Projects/UCLH-Foundry/PIXL/cli/src/pixl_cli/_config.py", line 21, in <module>
    "host": config("RABBITMQ_HOST"),
  File "/Users/milan/.pyenv/versions/pixl-3.10.13/lib/python3.10/site-packages/decouple.py", line 243, in __call__
    return self.config(*args, **kwargs)
  File "/Users/milan/.pyenv/versions/pixl-3.10.13/lib/python3.10/site-packages/decouple.py", line 105, in __call__
    return self.get(*args, **kwargs)
  File "/Users/milan/.pyenv/versions/pixl-3.10.13/lib/python3.10/site-packages/decouple.py", line 90, in get
    raise UndefinedValueError('{} not found. Declare it as envvar or define a default value.'.format(option))
decouple.UndefinedValueError: RABBITMQ_HOST not found. Declare it as envvar or define a default value.

```

milanmlft: So what I've done now is load the `.env` file from the current working directory with

```python
# Make sure the local .env file is loaded
load_dotenv(Path.cwd() / ".env")
```

so I think that's the second option you suggested?

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -53,7 +53,7 @@ def query_remote(self, data: dict, modality: str) -> Optional[str]:
         response = self._post(
             f"/modalities/{modality}/query",
             data=data,
-            timeout=config("PIXL_QUERY_TIMEOUT", default=10, cast=float),
+            timeout=config("PIXL_QUERY_TIMEOUT", default=120, cast=float),
```
stefpiatek, line 56: this feels quite long to find out if something exists on the VNA, why the change?

stefpiatek (approved):
🎉

pixl_imaging/src/pixl_imaging/_orthanc.py
milanmlft: Ah good point, I copied from what it is set to in `test/.env`:
https://github.com/UCLH-Foundry/PIXL/blob/2db63312b7aeeea6c38233a2a9ccf9d0f85857ea/test/.env#L4
🤷‍♂️

milanmlft: I'll revert it back to

```suggestion
            timeout=config("PIXL_QUERY_TIMEOUT", default=10, cast=float),
```

PR #313, opened 2024-02-21
Try build cache on passing imaging api tests

stefpiatek:
Default behaviour of github caching action is to not save to save if the workflow fails.

On passing test:
first run with caching enabled: [8m35s for entire imaging-api action](https://github.com/UCLH-Foundry/PIXL/actions/runs/7986327198?pr=313)
second run with caching enabled: [4m25s for entire imaging-api action](https://github.com/UCLH-Foundry/PIXL/actions/runs/7986432516?pr=313)

Updated the github caching action to always save cache

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -166,31 +166,6 @@ jobs:
     timeout-minutes: 10
     steps:
       - uses: actions/checkout@v3
-
-      - uses: docker/setup-buildx-action@v3
-      # pre-build and cache the postgres container as installing python3 takes a while, doesn't push
-      - name: Cache Docker layers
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -213,14 +188,15 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: docker/setup-buildx-action@v3
-      # pre-build and cache the orthanc-anon image as installing python3 takes a while, doesn't push
+      # pre-build and cache the postgres container as installing python3 takes a while, doesn't push
```
stefpiatek, line 191: I'll fix this in the next commit that tests the caching works on system test

PR #314, opened 2024-02-21
Correct instructions for editable installs

peshence:
We previously assumed `pip install -e` installed all following packages in editable mode, which is wrong.

### Old
```
pip install -e ../pixl_core .
```

which would copy the local package to the virtual environment, see pip freeze extract 

```
pixl_ehr @ file:///Users/petertsrunchev/Repositories/PIXL/PIXL/pixl_ehr
```
### New

```
pip install -e ../pixl_core -e .
```

which correctly installs local package as editable. 

Pip freeze:
```
-e git+ssh://git@github.com/UCLH-Foundry/PIXL.git@9c6f5a62fca5296d91cb4c93f647b9f037818086#egg=pixl_ehr&subdirectory=pixl_ehr
```

This updates the docs to reflect the correct way of installing editable packages.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -25,7 +25,7 @@ management tool such as [conda](https://docs.conda.io/en/latest/) or
 Then install in editable mode by running
 
 ```bash
-pip install -e ../pixl_core/ .
+pip install -e ../pixl_core/ -e .
 ```
 
 ## Usage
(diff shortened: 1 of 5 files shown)


milanmlft (approved):
🫶

stefpiatek (approved):
So much fun, thanks for getting that in

PR #324, opened 2024-02-27
Enable asynchronous consuming of messages

stefpiatek:
Had a great time, turns out the way we configured `pika-aio` was never switching to pick up the next message from the queue while in an `await` block. You can follow my fun debugging process but it now does what it was intended to do 🎉

The qos `set_qos(prefetch_count` argument sets the total number of messages which can be asynchronously processesed at once. Single threaded we average one message every 5 seconds, so 100 would allow for a maximum of 20 messages processed per second which allows us to use the token bucket to go to `12-15` with some wiggle room in case some message types take more


.env.sample
```diff
@@ -96,6 +96,7 @@ RABBITMQ_PASSWORD=
 # Imaging extraction API
 PIXL_DICOM_TRANSFER_TIMEOUT=240
 PIXL_QUERY_TIMEOUT=10
+PIXL_MAX_MESSAGES_IN_FLIGHT=100
```
milanmlft, line 99: Could we document this somewhere? 🙂

milanmlft (approved):
🚀 LGTM, just 1 suggestion to document what the `PIXL_MAX_MESSAGES_IN_FLIGHT` envvar means exactly, maybe in `pixl_core/README.md`?

peshence:
Looks great, for my understanding though - the async loop waits for each message to be processed before continuing to the next, while now if one message is being awaited, the next can start being processed?

stefpiatek:
> Looks great, for my understanding though - the async loop waits for each message to be processed before continuing to the next, while now if one message is being awaited, the next can start being processed?

Yep, an async for loop doesn't actually parallelise the iterable objects, it allows blocking iteration of the iterable

.env.sample
stefpiatek: done!

PR #329, opened 2024-03-01
Embed project name as tag in DICOM image

### LINKED ISSUE #296: Store project name in DICOM tags

- To enable routing of images to the correct study, use private DICOM tag to store:
  - [ ] project name (long string) 
  - Makes sense to do this from imaging-api after we've queried the VNA for the image (and if the image already exists in orthanc raw)
- [ ] In Orthanc anon. Use the project name to query for the `Export` row in pixl pipeline db, and then use that to find the `Image` that matches the DICOM data and the `Export` entity

### Testing

_No response_

### Documentation

_No response_

### Dependencies

_No response_

### Details and Comments

- project name is in the Message that `imaging-api` gets, and we should add it to the DICOM if the `Study exists in cache` or at the end of the `process_message` function in the imaging api's `_processing.py` file
- As a quick starter for ten, let's use the current tag for breed description because I think we're unlikely to want to keep that as it seems to be for non-humans. [(0010,2292)	LO	Patient Breed Description](https://www.dicomlibrary.com/dicom/dicom-tags/) 
- [Modification of a single instance](https://orthanc.uclouvain.be/book/users/anonymization.html#modification-of-a-single-instance) From anonymisation work you can supply the specific group and element tag so I think sending a rest request via the API like this should work
```
"Replace": {
              "0010-2292": "project-name"
            },
```

jeremyestein:
Fix #296 

TODO:
 - [x] Change group id from 0x000B to 000D?


orthanc/orthanc-anon/config/orthanc.json
```diff
@@ -1,4 +1,29 @@
 {
+  "Dictionary": {
+    "000b,1001": ["LO", "UCLHPIXLProjectName", 1, 1, "UCLH PIXL"]
+  },
```
jeremyestein, line 4: Surprisingly to me, removing this section seems to not affect the outcome - is this because we're never actually modifying this value though the API, it's just being allowed through unmodified during the anonymisation?
Should it be left in regardless?

projects/configs/tag-operations/test-extract-uclh-omop-cdm.yaml
```diff
@@ -591,3 +591,11 @@
   group: 0x7fe0
   element: 0x0010
   op: "keep" 
+- name: "Private Creator Data Element"
+  group: 0x000b
+  element: 0x0010
+  op: "keep"
+- name: "Private Creator Data Element"
```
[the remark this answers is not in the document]
stefpiatek: I think the names are just for human readability and not used in our code. But yeah I would have expected them to be different

orthanc/orthanc-raw/plugin/pixl.py
```diff
@@ -72,6 +79,53 @@ def should_auto_route():
...
+    # See the orthanc.json config file for where this tag is given a nickname
+    private_tag_offset = 0x01
+    # LO = Long string max 64, LT = long text max 10240, support paragraphs etc
+    # https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html
+    vr = "LO"
+    unknown_value = "__pixl_unknown_value__"
+    group_id = 0x000B
+    # The private block is the first free block >= 0x10. Other parts of the code assume
+    # it is == 0x10 though :/
+    # https://dicom.nema.org/dicom/2013/output/chtml/part05/sect_7.8.html
```
jeremyestein, line 117: @jstutters Did you say there was some solution for this? Were we going to just pick a higher value and hope it's free or something like that?

test/system_test.py
```diff
@@ -79,6 +80,29 @@ def two_zip_files_present() -> bool:
...
 
+        assert zip_files
+        for z in zip_files:
+            unzip_dir = tmp_path_factory.mktemp("unzip_dir", numbered=True)
+            self._check_dcm_tags_from_zip(z, unzip_dir)
+
+    def _check_dcm_tags_from_zip(self, zip_path, unzip_dir) -> None:
+        """
+        Check that private tag has survived anonymisation with the correct value.
+        Where are we testing patient name etc?
```
milanmlft, line 91: Who is this question addressed to? If it's something that still needs to be implemented, maybe create new issue?

milanmlft (approved):
Looking good! 🚀

orthanc/orthanc-anon/config/orthanc.json
```diff
@@ -1,4 +1,22 @@
...
+      "FrameOfReferenceUID",
+      "PatientID",
+      "PatientName",
+      "AccessionNumber",
+      "PixelSpacing",
+      "SpecificCharacterSet",
+      "BitsAllocated",
+      "BitsStored"
+    ]
+  },
```
stefpiatek, line 19: Oh what's the reasoning for having these here?

stefpiatek (changes requested):
We also need to use the project name from within orthanc anon (IIRC code where we do this is in pixl_dcmd), instead of querying the database to get the project name (see the checklist in the issue)

orthanc/orthanc-raw/plugin/pixl.py
jstutters: Yes, sadly the best we've got at the moment is to pick something high and unlikely to be selected for something else.  pydicom can search by name but the Orthanc modify method requires that the hex value be associated with a name in advance.  It should only be necessary to change the group ID.

orthanc/orthanc-anon/config/orthanc.json
jstutters: It makes the display in the UI a bit nicer but other than that serves no purpose in orthanc-anon.

orthanc/orthanc-raw/plugin/pixl.py
```diff
@@ -72,6 +75,53 @@ def should_auto_route():
...
+    Original from:
+    https://pydicom.github.io/pydicom/stable/auto_examples/memory_dataset.html
+    """
+    with BytesIO() as buffer:
+        dcmwrite(buffer, dataset)
+        buffer.seek(0)
+        return buffer.read()
+
+
+def modify_dicom_tags(receivedDicom: bytes, origin: str) -> Any:
```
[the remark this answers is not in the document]
jeremyestein: If I don't do it, it fails, but that's not to say there definitely isn't a way.

pixl_imaging/tests/docker-compose.yml
```diff
@@ -57,7 +57,8 @@ services:
       - "9001:8000"
 
   vna-qr:
-    image: osimis/orthanc:22.9.0-full-stable
+    image: orthancteam/orthanc:24.3.3
+    platform: linux/amd64
```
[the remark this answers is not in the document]
jeremyestein: It's what was done in the system test version of vna-qr, so I copied it from there.

test/system_test.py
jeremyestein: Yes, sorry, it seemed like a natural place to check all the DICOM metadata, and I don't think we do this anywhere else, so I wrote it down. It should go into an issue since it's not related to the project name feature.

orthanc/orthanc-anon/config/orthanc.json
jeremyestein: Well, it passes the system test without it. I can't remember if I was trying to fix something which I fixed another way, or thought it was needed for some other reason. This list was suggested somewhere in the orthanc docs ISTR. Will look into it.

orthanc/orthanc-raw/plugin/pixl.py
jeremyestein: I assume we're talking about the private block ID here? Ie. the BB in GGGG,BBOO.

This comment in the configuration file in the source code is interesting:
https://orthanc.uclouvain.be/hg/orthanc/file/Orthanc-1.12.3/OrthancServer/Resources/Configuration.json#l689

> Note: For private tags, you should only declare the lower 8 bits
> of the element since the higher 8 bits may vary from one file to
> the other.

I've tried setting the dictionary entry to "000b,01" and it doesn't immediately work, but this might be worth persuing as it's very much supposed to.

jeremyestein: Interestingly, the orthanc server is recognising the value as expected, even if it ends up in 0x1101. It's the API that's the problem.
image

orthanc/orthanc-raw/plugin/pixl.py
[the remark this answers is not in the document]
jeremyestein: It's possible that only the creation of the private block is required, rather than the tag itself.

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -76,6 +91,9 @@ def job_state(self, job_id: str) -> str:
     def _get(self, path: str) -> Any:
         return _deserialise(requests.get(f"{self._url}{path}", auth=self._auth, timeout=10))
 
+    def _delete(self, path: str) -> Any:
```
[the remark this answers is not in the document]
jeremyestein: I used it for testing but not any more

orthanc/orthanc-anon/config/orthanc.json
jeremyestein: That seems like a reason to keep it?

pixl_imaging/tests/docker-compose.yml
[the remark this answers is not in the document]
stefpiatek: Yeah I think also worth removing from there, feels strange to be limiting our platforms? Might be worth adding some M1-3 setup instructions with env variables to achieve this? @milanmlft or @peshence, what do you think?

orthanc/orthanc-raw/plugin/pixl.py
jeremyestein: pydicom is really not geared to the user explicitly chosing their block number, as far as I can tell. It's not supposed to matter but orthanc doesn't handle it right. Can we try using a higher group number (eg. 0x000d) and hoping we get block 0x10?

orthanc/orthanc-raw/plugin/pixl.py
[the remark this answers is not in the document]
jeremyestein: You can indeed just create the private block in the plugin, and the API will create the tag. You can even do nothing in the plugin, but the API will only create the data tag, but it'll have no private creator so the test fails.

pixl_imaging/tests/docker-compose.yml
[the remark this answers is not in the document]
peshence: I think I added this because of a warning that I thought at the time was a breaking error. I was wrong, it is not necessary.

orthanc/orthanc-anon/config/orthanc.json
stefpiatek: Yeah I'm fine with that

jeremyestein:
> We also need to use the project name from within orthanc anon (IIRC code where we do this is in pixl_dcmd), instead of querying the database to get the project name (see the checklist in the issue)

This is fixed now.

However, it's not ideal that the private tag info  (0x000B, 0x01, "UCLH PIXL", "UCLHPIXLProjectName") is copied over so many locations in python, json, and yaml.

The python locations could surely be easily reduced to one place, but the others may be a bit more fiddly.

One approach would be to have a single source of truth for our private tags in a json file, and use jsonnet to generate the various json config files from this. Python can trivially load the json file. For the yaml I'm not sure - yaml is a superset of JSON so maybe when loading the tag scheme we can concatenate it with the json file? Or just switch the tag schemes to json(net)?

Anyway, this should go in a separate issue. But I'll try and fix the python usages in this PR.

stefpiatek (approved):
Looks good, thanks Jeremy!

Can we also remove the use of `get_project_slug_from_hashid` from orthanc-anon/pixl.py, instead querying the API for the project name using the `resourceId`?, then we can pass that to the ftp upload, and then we don't have to query for the project name in `core.uploader._ftps` and then remove the `get_project_slug_from_hashid` entirely

```
hashed_patient_id = _get_patient_id(resourceId)
project_slug = get_project_slug_from_hashid(hashed_patient_id)
```

orthanc/orthanc-raw/plugin/pixl.py
```diff
@@ -72,6 +76,53 @@ def should_auto_route():
...
+    """
+    A new incoming DICOM file needs to have the project name private tag added here, so
+    that the API will later allow us to edit it.
+    However, we don't know its correct value at this point, so just create it with an obvious
+    placeholder value.
+    """
+    if origin != orthanc.InstanceOrigin.DICOM_PROTOCOL:
+        # don't keep resetting the tag values if this was triggered by an API call!
+        print("modify_dicom_tags - doing nothing as change triggered by API")  # noqa: T201
```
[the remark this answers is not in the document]
jeremyestein: I couldn't make it actually appear anywhere, am assuming orthanc does something with plugin stdout.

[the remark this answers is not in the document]
stefpiatek: fair enough, keep it as is for now. If it get a bit of time today I might make logging available

jeremyestein:
> Looks good, thanks Jeremy!
> 
> Can we also remove the use of `get_project_slug_from_hashid` from orthanc-anon/pixl.py, instead querying the API for the project name using the `resourceId`?, then we can pass that to the ftp upload, and then we don't have to query for the project name in `core.uploader._ftps` and then remove the `get_project_slug_from_hashid` entirely
> 
> ```
> hashed_patient_id = _get_patient_id(resourceId)
> project_slug = get_project_slug_from_hashid(hashed_patient_id)
> ```

Looking at `get_project_slug_from_hashid`, as a side effect it checks whether the image has already been exported.
Do we want to keep a call to `_query_existing_image` for this reason? Or will the uploader service fix this better anyway, and we can live with re-uploads for the time being? I'm not sure in what situations we would encounter this condition.

```
        existing_image = _query_existing_image(pixl_session, hashed_value)

        if existing_image.exported_at is not None:
            msg = "Image already exported"
            raise RuntimeError(msg)
```

orthanc/orthanc-raw/plugin/pixl.py
jeremyestein: I have moved it to group 0x000D and will mark this as resolved.

PR #333, opened 2024-03-04
Move export tests to PIXL core

milanmlft:
- Move `test_copy_omop.py` to `pixl_core` instead, as this is not testing any of the functionality in `cli`
- Move OMOP test resources to `pytest-pixl` for easy reuse in other tests


stefpiatek (approved):
🎉

PR #335, opened 2024-03-04
Refer to specific rows in the dataframe; don't assume order

jeremyestein:
This was failing on my dev machine.


milanmlft (approved):
👍

stefpiatek (approved):
fun!

PR #336, opened 2024-03-04
Make docker images smaller

jeremyestein:
Shave about 300MB off each of imaging-api, ehr-api, and fake cogstack.

This was inspired by out of disk errors in #329 , but whether it helps there or not it's a good thing anyway.

There is likely further scope for using multi-stage builds to make images smaller without the Dockerfiles getting too ugly, and reusing our base images.


milanmlft (approved):
🎉

PR #339, opened 2024-03-07
Update to latest orthanc containers with Python 3.11

milanmlft:
Using the new `orthancteam/orthanc` images, following their [update to Debian 12](https://github.com/orthanc-server/orthanc-builder/blob/master/release-notes-docker-images.md#2433).
This should allow us to use Python 3.11 everywhere.

### DIFF
--- a/test/docker-compose.yml
+++ b/test/docker-compose.yml
@@ -37,10 +37,15 @@ services:
         volumes:
             - ${PWD}/vna_config/:/run/secrets:ro
         healthcheck:
-            test: ["CMD", "curl", "-f", "-u", "orthanc:orthanc", "http://vna-qr:8042/patients"]
-            interval: 10s
-            timeout: 30s
-            retries: 5
+            test:
+                [
+                    "CMD-SHELL",
+                    "/probes/test-aliveness.py --user=orthanc, --pwd=orthanc",
+                ]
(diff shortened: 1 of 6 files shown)


peshence:
You've changed some time-related parameters here, would be good to explain why (it's currently unclear if it was even intentional)

milanmlft:
> You've changed some time-related parameters here, would be good to explain why (it's currently unclear if it was even intentional)

You mean the `healthcheck`s? It's recorded in the commit: 5316b6d2a49351eb55205732eb6a811043478172
But yeah, should probably add it to the PR description

test/docker-compose.yml
```diff
@@ -36,10 +37,15 @@ services:
...
-            retries: 5
+            test:
+                [
+                    "CMD-SHELL",
+                    "/probes/test-aliveness.py --user=orthanc --pwd=orthanc",
+                ]
+            start_period: 10s
+            retries: 2
+            interval: 3s
+            timeout: 2s
```
jeremyestein, line 48: Has this been tested on the GAE?
I investigated the use of the `start_` tags, but IIRC they only work on docker engine >=25.0 and the GAE uses 23.0 or something, so I think this will refuse to work on the GAE.

Issue #234, opened 2024-01-24
Do not run containers as root and do not create root owned files


jeremyestein:
### Definition of Done / Acceptance Criteria

* Containers are no longer run as root.
* Files created by containers on mounted file systems are no longer owned by root, and thus they are readable/writable/deletable by a normal user.

The criteria above will both be fixed if the container no longer runs as root. If for some reason you only implement a fix for the file permissions issue (eg. by chowning the files after creating them), this issue should be split so the running as root issue is not lost.

### Testing

System test should check which user each container is running as after bringing them up. Note that running on a non-linux host will falsely cause such a test to pass, as the special docker linux kernel on these platforms doesn't map container root to host root in the first place.

Any test that checks the presence and contents of an output file should also check the file's ownership. For an example, see `check_radiology_parquet.py` where these tests exist but have been temporarily demoted to logging only.

### Documentation

Depends on implementation as to whether user will have to do anything different because of this (eg. set an environment var)

### Dependencies

Some possible solutions would need a reconfiguration of the GAE via Atos, which could take some time.

### Rationale
* Running as root is a security hazard. If one of the services gets compromised then the attacker can do more bad stuff to the GAE than they would otherwise.
* Files owned by root cannot be deleted/moved/etc by a normal user (or indeed a system test).

### Possible implementations
* Run docker daemon in "rootless" mode - will require some changes via Atos?
* User namespaces (`userns_mode` ) - also requires admin?
* Set the user directly with the `user` property of the docker compose service spec.

milanmlft:
Yeahhh, I encountered this issue as well when trying to tear down the export directories created by tests for #226, which ran fine locally but then ran into errors on the GHA because indeed the Docker volumes were root-owned and so any files created there could not be deleted... I ended up working around this by [ignoring errors from `shutil.rmtree()`](https://github.com/UCLH-Foundry/PIXL/pull/226/commits/aa38ce203b14d1464cd17022b49cd8349746f16f#diff-cbdf7fb7063d5f09f680007b0f72b9fec2da518626601521500b2bba8ae4cd96R78), meaning the files won't actually get deleted on the GHA runner 🤷 

Turns out though that this is a [longstanding issue with Docker-compose named volumes](https://github.com/docker/compose/issues/3270). There are some workarounds but none of them are trivial.

stefpiatek:
Part of wider work on best practices on GAEs.

dram1964:
[...]

PR #345, opened 2024-03-14
Run as non-root user "pixl" where possible [force-system-test]

jeremyestein:
Partially addresses #234 .

TODO:
 - [ ] Will have to request a user (eg. "pixl") be added on the GAE with a known UID, and fix the UID of the container user "pixl" to match it.
 - [ ] Need to document that exports directory must be writable by user "pixl"


jeremyestein:
This has been replaced by #400

Issue #305, opened 2024-02-16
`mypy` and `ruff` do not detect as many errors when run through `pre-commit`


jeremyestein:
### Definition of Done / Acceptance Criteria

- [x] Mypy will detect common typing errors to an appropriate level as determined by the dev team (it is highly configurable)
- [x] Ditto ruff
- [x] Satisfactorily investigate why settings are different when run directly vs via pre-commit

### Details and Comments

I don't ever recall `mypy` failing in our pre-commit checks - is it just turned down to a very low level? Even the most obvious of typing errors isn't detected:
Quick test: add line `upload_parquet_files(42)` to `pixl_core/src/core/upload.py`.
Running `mypy pixl_core/src/core/upload.py` gives `error: Argument 1 to "upload_parquet_files" has incompatible type "int"; expected "ParquetExport"`. My IDE also gives the error.
But running `pre-commit run --all-files --verbose` passes just fine. I would expect it to fail too.

jeremyestein:
The problem seems to exist for `ruff` too.
Compare pre-commit:
```
$  pre-commit run --all-files --verbose
...
ruff.....................................................................Passed
- hook id: ruff
- duration: 0.03s
...
```

with running ruff directly:
```
$ ruff .
...
pixl_ehr/src/pixl_ehr/main.py:158:10: ASYNC101 Async functions should not call `open`, `time.sleep`, or `subprocess` methods
pixl_ehr/tests/test_processing.py:40:20: LOG002 Use `__name__` with `logging.getLogger()`
Found 2 errors.
...
```
Neither `LOG002` nor `ASYNC101` are explicitly mentioned in any config file.

milanmlft:
For `ruff` it's probably because we're using a hopelessly outdated version in pre-commit. `mypy` not sure yet, but likely something similar. Will open a PR soonish to update the `ruff` stuff and see if I can find a fix for `mypy`.

milanmlft:
OK so I got `ruff` fixed in #347 and did some updates for `mypy` as well.
BUT, `mypy` is still not catching much. E.g. the example of Jeremy still passes (while it shouldn't).

PR #347, opened 2024-03-18
Fix ruff in `precommit`

milanmlft:
- **Update deprecated ruff linter settings**
- **Update pre-commit-ruff**
- **ruff formatting and autofixes**
- **Ignore ruff ASYNC101 for async file opening**
- **Fix for ruff `LOG002`**

Related to #305

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,7 @@
 ---
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.1.6
+    rev: v0.3.3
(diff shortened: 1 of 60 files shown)


milanmlft:
`mypy` still doesn't capture everything it should.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [mypy]
```
stefpiatek, line 1: do we need both this and the mypy.ini? Feels like the same data

stefpiatek (approved):
Looks good, thanks for getting this in

Issue #292, opened 2024-02-12
Allow DICOM tag operations to be customised per MR model


stefpiatek:
### Definition of Done / Acceptance Criteria

priority may change based on investigations of MR machines used and their tags.

- [x] Validate and parse a configuration file per project. This has a way to define overrides for the default operations based on values of sepcific dicom tags 
  - [ ] Will be able to define a default action if none of the pre-defined values match the value in the current DICOM study. Would be good to have `do nothing` (so no overriding takes place) or `throw exception` (only known values are allowed)

### Testing

- [ ] module-level tests for overriding, doing nothing and throwing an exception

### Documentation

- [x] Update the documentation in orthanc anon

### Dependencies

- [x] #232 
- [ ] #291 

### Details and Comments

- Will need to be regex to match, 
- [ ] @stefpiatek to output list of options of manufacturer

stefpiatek:
DICOM tags to get the make and model:

Manufacturer (0008,0070)
ManufacturerModelName (0008,1090)

jstutters:
I wouldn't expect too much variation at the model level so ideally we'll be able to define operations at the manufacturer level then override those as necessary for models with different behaviour.

milanmlft:
### Plan

- [ ] Adapt `pytest_pixl.generate_dicom_dataset()` to add manufacturer tag(s) to test data
- [ ] Update `pixl_dcmd.merge_tag_schemes()` so it can actually merge multiple tag schemas
	- [ ] Should only be done if the dataset has the corresponding manufacturer tag!
- [ ] Create mock manufacturer-specific schema for testing, in `projects/configs/tag-operations/<manufacturer>.yaml`
- [ ] Decide what to do when a tag is specified in the schema, but is not present in the DICOM dataset

milanmlft:
Updating `pixl_dcmd.merge_tag_schemes()` to handle multiple tag schemes turns out to be a bit challenging. Because a single tag scheme read from a YAML file is a `list[dict]`, where `dict` has keys `['name', 'group', 'element', 'op']`. To merge two schemes, we would need to combine the two `list[dict]`s without duplicating some tags, which is tricky because there's no way to uniquely identify elements in a list, without checking the element itself.
One possible solution would be to convert the `list[dict]` to a `dict[dict]` and use the combination of `group` and `element`, which presumably are unique for each tag,  as the keys for the new `dict`.

PR #348, opened 2024-03-19
Parametrise `generate_dicom_dataset()`

milanmlft:
In preparation of #292

- **Add default DICOM tags as JSON file**
- **Parametrise `generate_dicom_dataset()`, enable customising any valid DICOM tag**


pytest-pixl/src/pytest_pixl/dicom.py
```diff
@@ -14,308 +14,107 @@
...
+    "instance_number": "1",
+    "image_position_patient": (76.0, -139.0, 119.0),
+    "slice_location": 82.0,
+    "window_centre": "321",
+    "window_width": "558",
+    "pixel_data": None,
+}
+
+
+def generate_dicom_dataset(tag_values: dict = TAGS_DICT, **kwargs) -> Dataset:
```
stefpiatek, line 63: Fun, would be good to document what the kwargs are doing in the defintion

stefpiatek (approved):
Thanks for getting this in, looks like should make it more customisable 🎉

pytest-pixl/src/pytest_pixl/dicom.py
```diff
@@ -14,308 +14,114 @@
...
-    ds.PerformedProcedureStepDescription = "Step Description"
-    ds.PerformedProcedureTypeDescription = ""
+    ds.PixelData = pixel_data.tobytes()
 
-    # Performed Protocol Code Sequence
-    performed_protocol_code_sequence = Sequence()
-    ds.PerformedProtocolCodeSequence = performed_protocol_code_sequence
+    # Handle any additional tags
+    for key, value in kwargs.items():
+        # Check if tag is DICOOM compliant
```
jstutters, line 104: Typo

pytest-pixl/src/pytest_pixl/dicom.py
```diff
@@ -14,308 +14,114 @@
...
 import json
 
 import numpy as np
-from pydicom.dataset import Dataset, FileMetaDataset
-from pydicom.sequence import Sequence
+from pydicom.datadict import dictionary_has_tag
+from pydicom.dataset import Dataset
 
 
 def write_volume(filename_pattern: str):
```
jstutters, line 27: Missing return type annotation (`-> None`)

pytest-pixl/src/pytest_pixl/dicom.py
```diff
@@ -14,308 +14,114 @@
...
-    ds.PerformedProcedureTypeDescription = ""
+    ds.PixelData = pixel_data.tobytes()
 
-    # Performed Protocol Code Sequence
-    performed_protocol_code_sequence = Sequence()
-    ds.PerformedProtocolCodeSequence = performed_protocol_code_sequence
+    # Handle any additional tags
+    for key, value in kwargs.items():
+        # Check if tag is DICOOM compliant
+        if dictionary_has_tag(key):
```
jstutters, line 105: I don't think this will work in practice.  For example, [this document](https://github.com/rordenlab/dcm2niix/tree/master/Philips#diffusion-direction) tells us that `(0x2005, 0x1412)` is important for diffusion images from a Philips scanner but that tag isn't present in pydicom's dictionary of known tags.  I'd expect the same to be true for other private space tags.

pytest-pixl/src/pytest_pixl/data/default_dicom_tags.json
jstutters, line 1: I think we're going to need some documentation on how to create one of these files e.g. from a real-world DICOM that we'd like to pass through PIXL.

jstutters (dismissed):
Looks a lot tidier than my approach.  Still need some work to handle private tags (unless I'm misunderstanding something).

pytest-pixl/src/pytest_pixl/dicom.py
milanmlft: Nice, clearly another indication that `mypy` is not working properly in the CI (#305) 🫠

pytest-pixl/src/pytest_pixl/dicom.py
milanmlft: I see, originally I didn't put this check in, but that allows setting any arbitrary tag name, as in `ds.blablbabla = "what"`. But I guess since this is just to generate test data, that's not such a big problem?

pytest-pixl/src/pytest_pixl/data/default_dicom_tags.json
milanmlft: Yeah you're right. I created this one essentially by running your original `generate_dicom_dataset()` and then exporting to JSON.

pytest-pixl/src/pytest_pixl/dicom.py
milanmlft: Or, how about having a separate function `add_private_tags()` that uses [`pydicom.Dataset.private_block()`](https://pydicom.github.io/pydicom/stable/old/private_data_elements.html#setting-private-data-elements-with-pydicom) to add private tags?

milanmlft:
Alright, so with `add_private_tags()`,  I think we can mimic the [Philips diffusion example](https://github.com/rordenlab/dcm2niix/tree/master/Philips#diffusion-direction):

```python
ds = generate_dicom_dataset(Manufacturer = "Philips")
add_private_tags(
    ds,
    [
        ((0x2001, 0x1003), "FL", 1000),
        ((0x2005, 0x10B0), "FL", 1.0),
        ((0x2005, 0x10B1), "FL", 1.0),
        ((0x2005, 0x10B2), "FL", 0.0),
    ],
)

```

pytest-pixl/README.md
```diff
@@ -16,4 +16,344 @@ the following environment variables:
...
+
+### Default DICOM tags
+
+[`default_dicom_tags.json`](./src/pytest_pixl/data/default_dicom_tags.json) contains a dictionary of
+default DICOM tag values used to generate fake DICOM data in `generate_dicom_dataset`.
+This JSON file was generated by manually creating a DICOM dataset with `pydicom` and then exporting
+to JSON with [`pydicom.Dataset.to_json_dict()`](https://pydicom.github.io/pydicom/stable/tutorials/dicom_json.html#converting-a-dataset-into-json-format).
+The details are given below.
+
+<details><summary>Default DICOM tags generation details</summary>
```
stefpiatek, line 95: This feels like we're duplicating what the code does, and would need to keep it up to date. What are you aiming for from this part?

stefpiatek (approved):
Looking good to me, not too sure about the benefit of having the code in documentation though

pytest-pixl/README.md
milanmlft: Because of [Jon's suggestion](https://github.com/UCLH-Foundry/PIXL/pull/348#discussion_r1532015939) to document how the `data/default_dicom_tags.json` file was created. This code doesn't actually appear anywhere. I ran it interactively to create the JSON file. So it's more of a provenance thing.

PR #349, opened 2024-03-20
Allow python 3.12 for cli

stefpiatek:
Don't want to have to recreate another set of envs on the GAE next time we upgrade. So allowing greater than 3.10


milanmlft:
I wonder what the reason was we restricted it 🤔

stefpiatek:
> I wonder what the reason was we restricted it 🤔

It was because of the pyyaml version not being able to be installed on 3.12, but we've fixed that now

milanmlft (approved):
🚀

PR #352, opened 2024-03-21
Allow merging multiple tag schemas

milanmlft:
In preparation of #292. So we can merge the manufacturer overrides with the base tag schema.

- **Add simple test for merging multiple tag schemes**
- **Implement merging multiple tag schemes**


peshence (approved):
🚀

PR #353, opened 2024-03-21
Allow running on GAE

stefpiatek:
Bit disjointed but a few changes to get current setup working

- Allow python 3.12 for CLI so we don't have to make another set of conda envs
- Tried to fight against async consuming blocking main thread (leading to rabbitmq heartbeat to fail and also restapi unresponsive after about 10 minutes of 0.2 messages per second) #355 
- Slimmed down logging for production
- For caching to work, all parameters must be hashable, so connected to azure inside private function
- Moved .secrets.env only to test dir, as that's where we use it
- Started using custom exception for when we want to print without stack trace 
- Really fun, maybe about a fifth of the time, the project name is in bytes when read from DICOM tag
- Updated CLI usage


pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -60,20 +61,26 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
...
     ).value
-    project_config = load_project_config(slug)
-    logger.error(f"Received instance for project {slug}")
+    # Get both strings and bytes, which is fun
+    if isinstance(raw_slug, bytes):
+        logger.debug(f"Bytes slug {raw_slug!r}")
+        slug = raw_slug.decode("utf-8").strip()
+    else:
+        logger.debug(f"String slug '{raw_slug}'")
+        slug = raw_slug
```
milanmlft, line 75: Is this a DICOM thing? 🫠

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -90,37 +97,25 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
     logger.info(
         f"DICOM tag anonymisation applied according to {project_config.tag_operation_files}"
     )
-    logger.warning("DICOM tag anonymisation applied")
-
     # Write anonymised instance to disk.
     return dataset
 
 
 def merge_tag_schemes(tag_operation_files: list[Path]) -> list[dict]:
```
milanmlft, line 104: Hmm, why did you revert the changes from #352? 
Will need this for #354

pixl_dcmd/tests/test_main.py
milanmlft, line 1: Also reverting #352

milanmlft (changes requested):
Looks good, but not sure why you reverted #352?

pixl_dcmd/src/pixl_dcmd/main.py
stefpiatek: Its certainly not documented, not quite sure why it happens only sometimes

milanmlft (approved):
👌

PR #354, opened 2024-03-21
Allow customisable tag operations per MR model

### LINKED ISSUE #292: Allow DICOM tag operations to be customised per MR model

milanmlft:
- **Add example tags for MRI diffusion data**
- **Update test config to take manufacturer overrides**
- **Update project config template with base and manufacturer overrides**
- **Update project config handler to allow manufacturer overrides**
- **Add simple test for merging multiple tag schemes**
- **project_config: make manufacturer overrides optional**
- **project_config: export `TagOperations`**
- **Refactor `merge_tag_schemes` so it can handle manufacturer overrides**
- **Update test for base-only tag scheme**
- **Add test fixture for manufacturer overridden tag scheme**
- **Test if manufacturer overrides work**
- **Actually apply the manufacturer overrides during anonymisation**

Fixes #292


pixl_core/src/core/project_config.py
```diff
@@ -60,6 +60,28 @@ class _Project(BaseModel):
     modalities: list[str]
 
 
+class TagOperations(BaseModel):
+    """Tag operations files for a project. At least a base file is required."""
+
+    base: Path
```
stefpiatek, line 66: I think this should probably be a list if we want to keep the same API for allowing merging of schemes open for the future

pixl_core/tests/test_project_config.py
```diff
@@ -30,16 +30,20 @@ def test_config_from_file():
...
-    ]
+    assert (
+        project_config.tag_operation_files.base
+        == PROJECT_CONFIGS_DIR / "tag-operations" / "test-extract-uclh-omop-cdm.yaml"
+    )
     assert project_config.destination.dicom == "ftps"
     assert project_config.destination.parquet == "ftps"
 
 
 BASE_YAML_DATA = {
```
stefpiatek, line 41: why aren't we just reading in the yaml file here? Feels like we'll always have to duplicate any changes

pixl_core/tests/test_project_config.py
```diff
@@ -67,6 +71,6 @@ def test_invalid_destinations():
 def test_invalid_paths():
     """Test that the config validation fails for invalid tag-operation paths."""
     config_data_wrong_base = BASE_YAML_DATA
```
stefpiatek, line 73: I expect you want to make a copy of this, or have the yaml data as a fixture otherwise if we tried to reuse it, the order of testing is important (would change items in the original `BASE_YAML_DATA` dict)?

pixl_core/tests/test_project_config.py
```diff
@@ -30,16 +30,20 @@ def test_config_from_file():
...
 
 BASE_YAML_DATA = {
     "project": {"name": "myproject", "modalities": ["DX", "CR"]},
-    "tag_operations": ["test-extract-uclh-omop-cdm.yaml"],
+    "tag_operations": {
+        "base": "test-extract-uclh-omop-cdm.yaml",
+        "manufacturer_overrides": "mri-diffusion.yaml",
+    },
     "destination": {"dicom": "ftps", "parquet": "ftps"},
 }
```
stefpiatek, line 49: Feels strange that we're not testing a load config method here for the overrides and that its in main instead

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -96,25 +96,42 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
...
+
+    manufacturer_overrides = _load_scheme(tag_operation_files.manufacturer_overrides)
+
+    # Keep only the overrides for the specified manufacturer
+    manufacturer_tags = [
+        tag
+        for override in manufacturer_overrides
+        if override["manufacturer"] == manufacturer
+        for tag in override["tags"]
+    ]
```
stefpiatek, line 117: FYI having manufacturer as the dict key would have made this easier, as we could just look up by dict returning empty list if not found. Fine to keep like this because we may end up needing to alter based on software version or model

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -96,25 +96,42 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
...
 
+def _load_scheme(tag_operation_file: Path) -> list[dict]:
+    with tag_operation_file.open() as file:
+        # Load tag operations scheme from YAML.
+        tags = yaml.safe_load(file)
+        if not isinstance(tags, list) or not all(
+            [isinstance(tag, dict) for tag in tags]
+        ):
+            raise ValueError("Tag operation file must contain a list of dictionaries")
+        return tags
```
stefpiatek, line 132: Should we just use pydantic? Then when merging we just call `model_dump()`

pixl_dcmd/tests/test_main.py
```diff
@@ -137,20 +139,87 @@ def test_can_nifti_convert_post_anonymisation(
...
+    tmpdir = tmp_path_factory.mktemp("tag-operations")
+    base_tags_path = tmpdir / "base.yaml"
+    with open(base_tags_path, "w") as f:
+        f.write(yaml.dump(base_tags))
+    manufacturer_overrides_path = tmpdir / "manufacturer_overrides.yaml"
+    with open(manufacturer_overrides_path, "w") as f:
+        f.write(yaml.dump(manufacturer_overrides_tags))
+
+    return TagOperations(
+        base=base_tags_path, manufacturer_overrides=manufacturer_overrides_path
```
stefpiatek, line 204: mentioned it before but seems strange that these are tests in main, where its testing a specific part of functionality rather than how users interact with dcmd. If we move out the `load_scheme` into a project config module that'd make more sense IMHO

projects/configs/uclh-nasogastric-tube-project-ngt-only-full-dataset.yaml
```diff
@@ -17,7 +17,7 @@ project:
     modalities: ["DX", "CR"]
 
 tag_operation_files:
-    ["uclh-nasogastric-tube-project-ngt-only-full-dataset.yaml"]
+    base: "uclh-nasogastric-tube-project-ngt-only-full-dataset.yaml"
```
stefpiatek, line 21: This will currently fail validation
```suggestion
    manufacturer_overrides: null
```

projects/configs/tag-operations/mri-diffusion.yaml
stefpiatek, line 1: is it worth having a subdirectory in tag operations of `modality` or `manufacturer`?

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -96,25 +96,42 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
...
+        if override["manufacturer"] == manufacturer
+        for tag in override["tags"]
+    ]
+
+    all_tags.update(_scheme_list_to_dict(manufacturer_tags))
 
     return list(all_tags.values())
 
 
+def _load_scheme(tag_operation_file: Path) -> list[dict]:
```
stefpiatek, line 124: feels like this should be in a config module, either in core or dcmd

stefpiatek (approved):
Thanks for working on this Milan, looks good - some suggestions here and there, happy to chat if morning me hasn't made much sense

pixl_core/tests/test_project_config.py
milanmlft: Move schema loading logic to `project_config.py`
Have wrapper class around `PixlConfig` and tag operations loaded as dictionaries

pixl_core/tests/test_project_config.py
milanmlft: Fixed in #356

milanmlft:
Started adding unit test for `anonymise_dicom()` but will take a bit more work, so deferring to #360.
For now, system test should catch any problems with the complete anonymisation.

pixl_core/src/core/project_config/tagoperations.py
stefpiatek, line 1: probably should use snake cake for the filename

pixl_core/src/core/project_config/pixlconfig_model.py
stefpiatek, line 1: why not `pixl_config_model`?

pixl_dcmd/src/pixl_dcmd/_tagschemes.py
```diff
@@ -0,0 +1,51 @@
...
+    tag_operations: TagOperations, manufacturer: Optional[str] = None
+) -> list[dict]:
+    """Merge multiple tag schemes into a single scheme."""
+    all_tags = {}
+
+    # Merge base tag schemes
+    for base_tags in tag_operations.base:
+        all_tags.update(_scheme_list_to_dict(base_tags))
+
+    if tag_operations.manufacturer_overrides is not None:
```
stefpiatek, line 34:
```suggestion
    if tag_operations.manufacturer_overrides:
```

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -66,7 +63,7 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
         DICOM_TAG_PROJECT_NAME.creator_string,
     ).value
     project_config = load_project_config(slug)
-    logger.error(f"Received instance for project {slug}")
+    logger.warning(f"Received instance for project {slug}")
```
stefpiatek, line 66: why warning?

pixl_dcmd/src/pixl_dcmd/_tagschemes.py
stefpiatek, line 1: also underscore missing here

pixl_dcmd/tests/test_tagschemes.py
stefpiatek, line 1: filename here as well

stefpiatek (approved):
Looking good, thanks for this Milan. Mostly some file naming suggestions

pixl_core/src/core/project_config/tagoperations.py
milanmlft: 🐍 🍰

pixl_dcmd/src/pixl_dcmd/main.py
milanmlft: It's more inline with the other logging messages. `error` feels weird, no?

milanmlft: I see that you've updated this to `logger.debug()` in `main`, which makes even more sense!

PR #356, opened 2024-03-22
Various fixes to config code and tests

jeremyestein:
- Optional fields are now truly optional
- Check tag op files exist
- Each test was repeatedly modifying the same test data; use a fixture instead
- Base data had some out of date field names; now validate the baseline data before trying to break anything
- Check that all production configs are valid

### DIFF
--- a/pixl_core/src/core/project_config.py
+++ b/pixl_core/src/core/project_config.py
@@ -56,7 +56,7 @@ def _load_and_validate(filename: Path) -> PixlConfig | Any:
 
 class _Project(BaseModel):
     name: str
-    azure_kv_alias: Optional[str]
+    azure_kv_alias: Optional[str] = None
     modalities: list[str]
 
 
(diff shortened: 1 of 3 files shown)


pixl_core/src/core/project_config.py
```diff
@@ -87,6 +87,7 @@ class PixlConfig(BaseModel):
     destination: _Destination
 
     @field_validator("tag_operation_files")
+    @classmethod
```
milanmlft, line 90:
```suggestion
```

Not strictly needed, as we define it in `pyproject.toml` under `classmethod-decorators`

milanmlft (approved):
Thanks Jeremy!

pixl_core/src/core/project_config.py
jeremyestein: I put it in because the pydantic docs recommend it, but eh

milanmlft: You're right, I was going off from an older version of pydantic (which we were using before). I got rid of the `pyproject.toml` line as we don't need it anymore and added the `@classmethod` to the other field validator as well.

PR #360, opened 2024-03-26
Add unit tests for anonymisation process

### LINKED ISSUE #132: Unit test for anonymisation process
## Definition of done

The anonymisation process needs a proper unit test.
https://github.com/UCLH-Foundry/PIXL/blob/0e6fd5665d844de6df54a0e772283ba9b8a30ae0/pixl_dcmd/src/pixl_dcmd/tests/test_pixl_dcmd.py#L125-L126

milanmlft:
Fixes #132

- **Refactor: move `add_private_tag` to `core.dicom_tags`**
- **Only load project config when it's needed**
- **Add note to refactor private tag addition in `orthanc-raw`**
- **Add DICOM image test fixtures to test anonymisation**
- **Add anonymisation unit tests**


peshence (approved):
The whole private tag class is very odd to me, but that was already in there and everything else looks good. 🚀

milanmlft:
> The whole private tag class is very odd to me, but that was already in there and everything else looks good. 🚀

What do you think is odd about it? 
I guess it's just meant as a consistent interface to add custom private tags to DICOM datasets.

pixl_dcmd/tests/test_main.py
```diff
@@ -53,8 +126,49 @@ def test_remove_overlay_plane() -> None:
...
+def test_anonymisation(row_for_dicom_testing, vanilla_dicom_image: Dataset) -> None:
+    """
+    Test whether anonymisation works as expected on a vanilla DICOM dataset
+    """
+
+    orig_patient_id = vanilla_dicom_image.PatientID
+    orig_patient_name = vanilla_dicom_image.PatientName
+
+    # Sanity check: study date should be present before anonymisation
+    assert vanilla_dicom_image.StudyDate is not None
```
stefpiatek, line 138: Should this be in this form so that it matches our later assert?
```suggestion
    assert "StudyDate" in vanilla_dicom_image
```

stefpiatek (approved):
Nice, thanks for getting this in

Issue #359, opened 2024-03-26
Merge `pixl_dcmd` into `pixl_core`


jeremyestein:
From `pixl_dcmd/README.md`:
> The reason for having this as a separate package instead of having the functionality in `pixl_core` is because `orthanc` requires Python 3.9, whereas the rest of PIXL is on 3.10 or higher.

Since we're now on 3.11 for everything, I think pixl_dcmd can be merged into pixl_core, to reduce the number of packages we have and make things a bit easier to work with.

milanmlft:
Note, from #361, it looks like there is some clash between `pixl_dcmd/tests/conftest.py` and `hasher/tests/conftest.py`. 
Might be worth resolving this together with the merge.

jeremyestein:
If you're refactoring this area anyway, you might consider moving some of the not publicly called methods in `pixl_dcmd/src/pixl_dcmd/main.py` into one of the private modules adjacent to it.

PR #361, opened 2024-03-27
Allow running (almost) all module tests from root

milanmlft:
Minor QOL improvement, add a `pytest.ini` allowing to run `pytest` from root and run all module-level tests at once.
Note there might be some clashes in the `docker` services being spun up though.

Also, `pixl_dcmd/tests` and `hasher/tests` have some clashing `conftest` (though haven't pinned it down), so can only add one of the two.


stefpiatek (approved):
Fun with the clashing, expect we'll have to resolve this when we move dcmd into core, but we can get to that later on. Happy for you to merge and make an issue for that

milanmlft:
> Fun with the clashing, expect we'll have to resolve this when we move dcmd into core, but we can get to that later on. Happy for you to merge and make an issue for that

Tracking as part of https://github.com/UCLH-Foundry/PIXL/issues/359#issuecomment-2036772747

PR #365, opened 2024-04-09
Force reinstallation of PIXL modules in Docker containers

milanmlft:
🐳

### DIFF
--- a/docker/.dockerignore
+++ b/docker/.dockerignore
@@ -1,0 +1,2 @@
+build
+**/*/build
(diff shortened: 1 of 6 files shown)


peshence (commented):
Nice, but before we copy the actual code over we install just the dependencies, shouldn't that still be cached?

milanmlft:
> Nice, but before we copy the actual code over we install just the dependencies, shouldn't that still be cached?

I think it doesn't hurt. The `--no-cache-dir` was actually a suggestion from my Docker linter, so it seems more of a good practice thing.

stefpiatek (approved):
:shipit:

Issue #291, opened 2024-02-09
Use well tested anonymisation library instead of our own for DICOM tags


stefpiatek:
### Definition of Done / Acceptance Criteria

- [x] Create an options analysis for libraries which can be used to carry out tag operations which can 
  - [x] read in a config file/dictionary for tag operations
  - [x] parse datetimes and shift them
  - [x] remove tag
  - [x] replace tag with empty value
  - [x] apply a custom function to the tag 
- [ ] Document options analysis in PIXL repo
- From [options analysis](https://uclh.slab.com/posts/dicom-tag-processing-1cdo30rq#h9d4r-recommendations) we've decided on https://github.com/KitwareMedical/dicom-anonymizer
  - [ ] Use the library instead of our custom code, may require restructuring of the existing tag operations yaml file

### Documentation

- [ ] Update orthanc anon documentation 

### Details and Comments

Currently we're rolling our own DICOM tag operations which is error prone and I'd expect exists elsewhere

jstutters:
One for consideration: https://github.com/KitwareMedical/dicom-anonymizer

@HChughtai mentioned something from the flywheel.io team.

HChughtai:
I was just looking at it and checking if it was still active (which I think it is) - https://flywheel-io.gitlab.io/public/migration-toolkit/

My plan is to find a few popular options and look into them in more detail. So more options are good!

The other ones I've come across
- https://pydicom.github.io/deid/
- https://mircwiki.rsna.org/index.php?title=MIRC_CTP (not a Python option)

stefpiatek:
Need to work on decision on whether we continue with plugin based or more scalable open source options

HChughtai:
Writing up my options analysis on Slab at [DICOM Tag Processing](https://uclh.slab.com/posts/dicom-tag-processing-1cdo30rq)

HChughtai:
Options analysis now has some [broad recommendations](https://uclh.slab.com/posts/dicom-tag-processing-1cdo30rq#h9d4r-recommendations) that I'll chat with @stefpiatek about this week

> The three Python libraries considered have broadly the same feature sets. However the Flywheel-Migration library seems to be less supported compared to the Kitware Medical dicom-anonymizer and Pydicom deid. If we choose to go with either of the latter two then the existing Orthanc-based approach can be retained.
> 
> Alternatively we may wish to take this opportunity to rearchitect the orthanc-anon component to fit current and expected new requirements.

Next steps are:

[...]

PR #366, opened 2024-04-10
feat: 291 switch to third party anonymisation

peshence:
#291 Switch from DIY DICOM anonymiasation to [Kitware DICOM Anonymizer](https://github.com/KitwareMedical/dicom-anonymizer). In the process:

- Public DICOM tags now have default actions as per the library.
- old "hash" function applied to UIDs while keeping their prefix has been replaced with a random UID replacement (no prefix continuity) as we don't currently believe loading images on original scanner is likely.
- only patientID is now hashed via hasher api (as patientID+Acc_num). Previously Patient name and acc_num were also hased like this, now they are replaced as empty/constant strings in existing projects since they shouldn't be used for linking.
- removed date shifting options
- anonymisation and whitelisting now handle sequences

Possible improvement:
Convert current YAMl tag configs to JSON to make use the the existing underlying action loading functionality. We can then do regex replacements and a few other slightly more complex actions. Not sure if it's necessary at this stage.

Possible risks:
Private tags may change element number (from 0x10xx to 0x11xx e.g.) and we would not action on them (but they should be deleted by whitelist). There is also a very slight possibility in the future a tag has a keep action, but as another private creator existed upstream we keep the wrong tag, which may contain private data. Mitigation would be to specify private tags properly in our operations (so group, tag, creator) and then adjusting the tag actions accordingly for each dataset.


pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -82,344 +83,150 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
...
+
+    logger.warning("Anonymising received instance")
 
+    anonymize_dataset(dataset, tag_actions, delete_private_tags=False)
 
-def remove_overlays(dataset: Dataset) -> Dataset:
+    for elem in dataset.iterall():
+        if elem.VR == "SQ":
+            for sq_el in elem.value:
+                _anonymise_dicom_recursively(sq_el, modalities, tag_actions)
```
stefpiatek, line 132: 🎉 

is it worth setting a maximum depth as we're doing something recursive?

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -82,344 +83,150 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
...
             del dataset[del_grp, del_el]
             message = "Whitelist - deleting: {name} (0x{grp:04x},0x{el:04x})".format(
                 name=de.keyword, grp=del_grp, el=del_el
             )
             logger.debug(f"\t{message}")
-
-    return dataset
-
-
-def apply_tag_scheme(dataset: Dataset, tags: list[dict]) -> Dataset:
```
stefpiatek, line 166: this is so lovely to see

stefpiatek (commented):
Wasn't sure if you wanted a review but had a bit of time to have a look.

pixl_dcmd/tests/test_main.py
```diff
@@ -160,16 +160,16 @@ def test_anonymisation_with_overrides(
...
+    anonymise_dicom(mri_diffusion_dicom_image)
 
     # Whitelisted tags should still be present
-    assert (0x0010, 0x0020) in anon_ds
-    assert (0x2001, 0x1003) in anon_ds
-    assert anon_ds.PatientID != original_patient_id
+    assert (0x0010, 0x0020) in mri_diffusion_dicom_image
+    assert (0x2001, 0x1003) in mri_diffusion_dicom_image
+    assert mri_diffusion_dicom_image.PatientID != original_patient_id
     assert mri_diffusion_dicom_image[(0x2001, 0x1003)] == original_private_tag
```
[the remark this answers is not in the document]
peshence: we don't, I made this pr to see if system tests were failing locally only (they were), and forgot to tag it as such - my bad, but thanks for the review anyway!

I rewrote these tests to create proper private tags and valid dicom and thrown myself in for a loop a bit, will update later

projects/configs/tag-operations/uclh-nasogastric-tube-project-ngt-only-full-dataset.yaml
```diff
@@ -582,20 +582,12 @@
...
 - name: "Pregnancy Status"
   group: 0x0010
   element: 0x21c0
-  op: "delete" 
+  op: "delete"
 - name: "Pixel Data"
   group: 0x7fe0
   element: 0x0010
-  op: "keep" 
-- name: "Private Creator Data Element"
```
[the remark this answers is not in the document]
peshence: This was not intentional, good spot!

pixl_dcmd/tests/test_main.py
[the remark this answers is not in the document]
peshence: tests added

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -81,344 +81,139 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
...
-        f"DICOM tag anonymisation applied according to {project_config.tag_operation_files}"
+        f"Applying DICOM tag anonymisation according to {project_config.tag_operation_files}"
     )
-    # Write anonymised instance to disk.
-    return dataset
+    logger.debug(f"Tag scheme: {tag_scheme}")
+
+    if (0x0008, 0x0060) in dataset and dataset.Modality not in modalities:
+        msg = f"Dropping DICOM Modality: {dataset.Modality}"
+        logger.error(msg)
```
stefpiatek, line 97: feels like we should just raise the error instead of logging and raising?

peshence:
This is now ready for review. I have also updated the description, repeating risk here:
> Possible risks:
> Private tags may change element number (from 0x10xx to 0x11xx e.g.) and we would not action on them (but they should be deleted by whitelist). There is also a very slight possibility in the future a tag has a keep action, but as another private creator existed upstream we keep the wrong tag, which may contain

Should I make a ticket or is this not something we are really worried about?

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -81,344 +81,139 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
...
+        raise ValueError(msg)
```
stefpiatek, line 98: Might be worth using a custom Exception type could rename `PixlSkipMessageError` to `PixlDiscardError` or similar?

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -81,344 +81,139 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
...
 
+    logger.warning("Anonymising received instance")
```
stefpiatek, line 100: Should be an info level, though I know we've done terrible things to make it visible from the orthanc plugin. 

Now that we've got loguru in core (used core, cli and pixl_imaging) we can do some simple logging config in the `pixl.py` plugin, the fun thing is that it only prints stdout not stderr, but I think this should get the orthanc-anon to print the logs as we'd expect. 

```
# Set up logging as main entry point
logger.remove()  # Remove all handlers added so far, including the default one.
logging_level = config("LOG_LEVEL")
if not logging_level:
    logging_level = "INFO"
logger.add(sys.stdout, level=logging_level)

logger.warning("Running logging at level {}", logging_level)

```

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -81,344 +81,139 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
...
+            add_hashed_identifier_and_save_to_db(existing_image, hashed_value)
+        elif dataset[grp, el].VR == "SH":
+            pat_value = str(dataset[grp, el].value)
+            hashed_value = _hash_values(pat_value, 16)
 
+        dataset[grp, el].value = hashed_value
+
+    else:
+        message = f"Missing linking variable (0x{grp:04x},0x{el:04x})"
+        logger.warning(f"\t{message}")
```
stefpiatek, line 181: this happens quite a lot, I wonder if it should also be debug

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -81,344 +81,139 @@ def anonymise_dicom(dataset: Dataset) -> Dataset:
...
+    """
+    dataset.walk(lambda ds, de: _whitelist_tag(ds, de, tag_scheme), recursive)
+
+
+def _whitelist_tag(dataset: Dataset, de: DataElement, tag_scheme: list[dict]) -> None:
+    """Delete element if it is not in the tagging schemе."""
+    tag_dict = _scheme_list_to_dict(tag_scheme)
+    if (
+        # de.VR != "SQ"
+        # and (de.tag.group, de.tag.element) in tag_dict
```
stefpiatek, line 214: should probably just remove this, interesting that ruff didn't complain about it

pixl_dcmd/src/pixl_dcmd/__init__.py
```diff
@@ -16,18 +16,12 @@
...
-__all__ = [
-    "remove_overlays",
-    "write_dataset_to_bytes",
-    "apply_tag_scheme",
-    "enforce_whitelist",
-]
+__all__ = ["anonymise_dicom", "write_dataset_to_bytes", "should_exclude_series"]
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.INFO)
```
stefpiatek, line 27: ah logging, `pixl_dcmd` is a library so shouldn't really be setting the logging level, commented later down the line but we should set the logging level in the `pixl.py` of orthanc raw, and replace the logging within pixl_dcmd to use `loguru` 
this should be the only interaction with the logger (other than logging) that we do in `dcmd`
```
from loguru import logger
```

stefpiatek (approved):
Looking good, appreciate your effort on this - know its been a bit more painful than we all expected. Mostly comments about tidying up logging, but we should fix this so we can use appropriate logging levels

pixl_dcmd/src/pixl_dcmd/main.py
peshence: does it? This will only be activated if there is no patient id currently

pixl_dcmd/src/pixl_dcmd/main.py
peshence: oh good catch

Issue #331, opened 2024-03-04
Running as a production service


stefpiatek:
### Definition of Done / Acceptance Criteria

- [x] UCLH-Foundry/PIXL#238
- [ ] UCLH-Foundry/PIXL#153
- [x] UCLH-Foundry/PIXL#305
- [ ] UCLH-Foundry/PIXL#193
- [ ] UCLH-Foundry/PIXL#310
- [ ] UCLH-Foundry/PIXL#290
- [x] UCLH-Foundry/PIXL#236
- [ ] UCLH-Foundry/PIXL#369
- [ ] UCLH-Foundry/PIXL#388
- [ ] UCLH-Foundry/PIXL#395
- [ ] UCLH-Foundry/PIXL#398

PR #367, opened 2024-04-10
Pytest classes cannot have an __init__ method

jeremyestein:
Several tests were being ignored in the system test.

### DIFF
--- a/test/system_test.py
+++ b/test/system_test.py
@@ -28,13 +28,12 @@ pytest_plugins = "pytest_pixl"
 class TestFtpsUpload:
     """tests adapted from ./scripts/check_ftps_upload.py"""
 
-    def __init__(self) -> None:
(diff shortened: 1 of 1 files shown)


stefpiatek (approved):
Ha yep that'd do it, frustrating that pytest fails silently here

jeremyestein:
> Ha yep that'd do it, frustrating that pytest fails silently here

One reason I'd like to do some mutation testing one day! If you change a load of your code and your tests still pass, your tests are probably broken.
A more down to earth measure might be to track test coverage over time so we can investigate any sudden drops.

milanmlft:
> A more down to earth measure might be to track test coverage over time so we can investigate any sudden drops.
Maybe worth adding to #331?

Also, sorry for this! This was part of fixing the `mypy` errors for #350, but did not realise the tests were failing 🙈

PR #370, opened 2024-04-11
ehr-api becomes export-api; losing and gaining some features

### LINKED ISSUE #321: Convert ehr api to export api

Once we have finished off the 100 days integration of new model for redaction in UCLH-Foundry/the-rolling-skeleton/issues/214 , strip out the emap star querying from PIXL

- [x] Remove tables from PIXL DB relating to EHR data (this is now in the OMOP extract)
- [x] Convert ehr api to export api and rename endpoints for exporting data (see top level document for step naming)
- [x] Remove redundant tests, fake cogstack api, fake emap star etc.
- [x] the pixl CLI create radiology linker table on the fly (and writes to export dir): on export will read in the omop data, query the pixl pipeline database for all images in the project name, and merge the omop data with the hashed imaging identifier to write the linker parquet file
- [ ] Rename EHR env vars and document that these need renaming in config file
- [x] Remove ehr queue
- [x] move orthanc anon's ftps export into the export api. orthanc anon tells export api that this resource is ready for export

### Testing

_No response_

### Documentation

- [ ] Update documentation to reflect exporting function, documenting pathways for exporting images and parquet files, imaging alone and parquet alone?

### Dependencies

- [ ] UCLH-Foundry/the-rolling-skeleton/issues/214

### Details and Comments

_No response_

jeremyestein:
Fixes #321


milanmlft:
Will need a few extra tweaks when #379 is merged in, but happy to help out with that!

jeremyestein:
> Will need a few extra tweaks when #379 is merged in, but happy to help out with that!

The merge wasn't too bad and the system test passes for me. Just fixing the core tests, but you can probably start checking that our bits are all joined up.

cli/src/pixl_cli/_config.py
```diff
@@ -54,7 +54,7 @@ def base_url(self) -> str:
 
 
 API_CONFIGS = {
-    "ehr_api": APIConfig(
+    "export_api": APIConfig(
         host=config("PIXL_EHR_API_HOST"),
         port=int(config("PIXL_EHR_API_PORT")),
         default_rate=float(config("PIXL_EHR_API_RATE", default=1)),
```
milanmlft, line 60: Should these `PIXL_EHR_API_*` variables also be renamed?

pixl_export/src/pixl_export/main.py
```diff
@@ -0,0 +1,110 @@
...
+    study_id = study_data.study_id
+    hashed_image_id, project_slug = get_tags_by_study(study_id)
+    project_config = load_project_config(project_slug)
+    destination = project_config.destination.dicom
+
+    uploader = get_uploader(project_slug, destination, project_config.project.azure_kv_alias)
+    msg = f"Sending {study_id} via '{destination}'"
+    logger.debug(msg)
+    zip_content = get_study_zip_archive(study_id)
+    uploader.upload_dicom_image(zip_content, hashed_image_id, project_slug)
```
milanmlft, line 110: This won't work for the `DicomWebUploader`, as it requires just a resource ID (which is why `DicomWebUploader.upload_dicom_image()` currently throws `NotImplementedError` and we have `DicomWebUploader.send_via_stow()`.

I wonder whether we could move the `get_study_zip_archive(study_id)` to `FTPSUploader.upload_dicom_image()` and then make all `upload_dicom_image()` methods accept only a `resource_id` parameter?

pixl_export/src/pixl_export/main.py
```diff
@@ -0,0 +1,110 @@
...
+import logging
+from datetime import (
+    datetime,  # noqa: TCH003, always import datetime otherwise pydantic throws error
+)
+from pathlib import Path
+
+from core.exports import ParquetExport
+from core.project_config import load_project_config
+from core.rest_api.router import router
+from core.uploader import get_uploader
```
milanmlft, line 28: Since `core.uploader` is now only used by `pixl_export` should we just move it over from `core`?

milanmlft (commented):
Looking great! 🚀 
I think we just need to tweak the `Uploader.upload_dicom_image()` method a bit and make sure the Dicomweb uploader is still working.

pixl_imaging/README.md
```diff
@@ -2,7 +2,7 @@
 
 The PIXL imaging API processes messages from the imaging queue created by the [CLI](../cli/README.md) 
 to query images from the [VNA](https://en.wikipedia.org/wiki/Vendor_Neutral_Archive) and transfers them to the [`orthanc-raw` instance](../orthanc/orthanc-raw/README.md).
-It is identical in structure the [EHR API](../pixl_ehr/README.md).
+It is identical in structure the [EHR API](../pixl_export/README.md).
```
peshence:
```suggestion
It is identical in structure the [Export API](../pixl_export/README.md).
```

test/docker-compose.yml
```diff
@@ -44,39 +44,6 @@ services:
...
-        build:
-            context: ./dummy-services/cogstack/
-        healthcheck:
-            test: ["CMD", "curl", "-f", "http://localhost:8000/heart-beat"]
-            interval: 10s
-            timeout: 30s
-            retries: 5
-        networks:
-            pixl-net:
-
```
peshence, line 79: ❤️

milanmlft:
OK so I pushed some changes to update the DICOMweb uploader and its tests, which includes a rework done by @stefpiatek in #378. This will break `core` and system tests, but that's to be expected because I don't think the `DicomWebUploader` is working... Sorry about this, #379 shouldn't have been merged in after all 😞

peshence (approved):
Looks good, mostly pointing out some docs that potentially need updating, feel free to resolve all of these as you see fit.

Nice work! 🚀

cli/src/pixl_cli/main.py
```diff
@@ -146,20 +149,25 @@ def populate(
...
 )
-def extract_radiology_reports(parquet_dir: Path) -> None:
+def export_patient_data(parquet_dir: Path) -> None:
     """
     Export processed radiology reports to parquet file.
 
-    PARQUET_DIR: Directory containing the extract_summary.json
-    log file defining which extract to export radiology reports for.
+    PARQUET_DIR: Directory containing the extract_summary.json log file
+                 defining which extract to export patient data for. (not a CSV)
```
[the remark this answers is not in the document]
stefpiatek: I think this is because you don't get any patient data from a csv input project.

pixl_core/src/core/uploader/_dicomweb.py
```diff
@@ -36,13 +36,15 @@ def __init__(self, project_slug: str, keyvault_alias: Optional[str]) -> None:
     def _set_config(self) -> None:
         # Use the Azure KV alias as prefix if it exists, otherwise use the project name
         az_prefix = self.keyvault_alias
-        az_prefix = az_prefix if az_prefix else self.project_slug
+        self.az_prefix = az_prefix if az_prefix else self.project_slug
```
stefpiatek, line 39: oh hello, fine for you to put this in as fixing it

pixl_dcmd/src/pixl_dcmd/_database.py
```diff
@@ -34,13 +35,13 @@
...
     PixlSession = sessionmaker(engine)
     with PixlSession() as pixl_session, pixl_session.begin():
         existing_image.hashed_identifier = hashed_value
         pixl_session.add(existing_image)
 
-        updated_image: Image = (
+        updated_image: Optional[Image] = (
             pixl_session.query(Image)
             .filter(
                 Image.accession_number == existing_image.accession_number,
```
stefpiatek, line 47: Been a while since I thought about this, but why are we querying for one or none here? Can there be a time where we don't find an image by the identifiers?

pixl_export/src/pixl_export/main.py
```diff
@@ -0,0 +1,110 @@
...
+
+app = FastAPI(
+    title="export-api",
+    description="Export service",
+    version=importlib.metadata.version("pixl_export"),
+    default_response_class=JSONResponse,
+)
+app.include_router(router)
+
+logger = logging.getLogger("uvicorn")
```
stefpiatek, line 43: May have mentioned elsewhere but worth switching over fully to using loguru for export

pixl_export/src/pixl_export/main.py
stefpiatek: Yeah I think this is a good shout, then we can use the same `upload_dicom_image` for both

pixl_export/tests/test_export_processing.py
```diff
@@ -0,0 +1,58 @@
...
+    content: str | None
+    text: str | None
+
+
+@pytest.fixture(autouse=True)
+def _mock_requests(monkeypatch) -> None:
+    """Mock requests so we don't have to run APIs."""
+
+    def mock_get(url: str, params: dict, *args: Any, **kwargs: Any) -> MockResponse:
+        logger.info("Mocking request for %s: %s", url, params)
```
stefpiatek, line 48: also when changing to loguru, will need to use `{}` style formatting instead of `%`

test/docker-compose.yml
```diff
@@ -44,46 +44,13 @@ services:
             timeout: 2s
         networks:
             pixl-net:
-
-    star:
```
stefpiatek, line 48: ah so nice

cli/src/pixl_cli/main.py
```diff
@@ -146,20 +149,25 @@ def populate(
...
+    linker_data = make_radiology_linker_table(parquet_dir, images)
+    export.export_radiology_linker(linker_data)
+
+    # Call the Export API
+    api_config = api_config_for_queue("export")
     response = requests.post(
         url=f"{api_config.base_url}/export-patient-data",
-        json={"project_name": project_name, "extract_datetime": omop_es_datetime.isoformat()},
+        json={"project_name": project_name_raw, "extract_datetime": omop_es_datetime.isoformat()},
         timeout=300,
```
stefpiatek, line 171: we should make this timeout configurable using the CLI I think

stefpiatek (approved):
Really nice to see us stripping away so much. Thanks for getting this in Jeremy. Some suggestions, think integration of dicomweb will take a bit and then worth switching over to loguru

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -219,18 +181,13 @@ def should_auto_route() -> bool:
 
 def _azure_available() -> bool:
     # Check if AZ_DICOM_ENDPOINT_CLIENT_ID is set
-    return config("AZ_DICOM_ENDPOINT_CLIENT_ID", default="") != ""
+    return bool(config("AZ_DICOM_ENDPOINT_CLIENT_ID", default="") != "")
```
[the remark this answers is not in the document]
jeremyestein: At one point it gave me the following mypy error:
`error: Returning Any from function declared to return "bool"  [no-any-return]`
The underlying reason is that `__eq__` doesn't have to return a bool in python (in numpy it could be an array of bools for example).
However, I can only reproduce this error when calling `mypy` explicitly on that file; it doesn't break our `pre-commit` checks or show up in my IDE, so I don't know why I ever encountered it. I think our `mypy` config is still a bit weird.
I will remove it because I don't like it and it seems to not break anything any more.

cli/src/pixl_cli/main.py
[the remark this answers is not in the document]
jeremyestein: It was to contrast it with `populate`, which can once again accept CSV files. That surprised me and I wanted to check that export_patient_data couldn't too. I found the answer and put it there to save someone else the time. But if it's causing more confusion then help I'll remove it!

pixl_export/src/pixl_export/main.py
jeremyestein: I tried this but it's complicated by `ParquetExport` (currently in core) using parts of `core.uploader`. We could move `ParquetExport` to export-api as well, but then we run into the fact that `ParquetExport` is used by both export-api and CLI.
I tried moving various bits around and got into a circular import hell. I am probably missing some obvious solution, but it could be that the fact we have CLI doing exports makes this inherently harder than it needs to be. Maybe worth a separate issue/PR?

pixl_export/src/pixl_export/main.py
jeremyestein: Resolving this as I assume it's addressed in #381

pixl_export/src/pixl_export/main.py
jeremyestein: Now issue #384

cli/src/pixl_cli/main.py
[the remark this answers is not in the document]
jeremyestein: It got re-added after having been deleted, not sure why.
I'll remove the comment above.

pixl_dcmd/src/pixl_dcmd/_database.py
jeremyestein: I don't know what this code is trying to achieve. It seems to be querying the DB for the row that it just added/updated. The return value is not used anyway in the one place this method is called. So unless the query has a side-effect it could probably be deleted altogether.
Maybe we need a separate task for tidying up our usage of sqlalchemy, since the session creation is also a bit copy n pastey?

pixl_core/src/core/uploader/_dicomweb.py
jeremyestein: This has been reverted and is now being fixed in #381

pixl_dcmd/src/pixl_dcmd/_database.py
jeremyestein: Now issue #385

PR #371, opened 2024-04-11
Add per-project hashing salt

### LINKED ISSUE #343: Per project salt in PIXL hashing

- [ ] PIXL hasher uses the configured project yaml files to have a project-specific salt

### Testing

_No response_

### Documentation

_No response_

### Dependencies

- Anonymisation library work

### Details and Comments

We're now keeping the PIXL-specific hashing API for this iteration so need to make sure that we use a different salt per project in hashing

Suggested flow

- [ ] Use project name as the key in azure keyvault, which stores the salt (note: separate Azure Keyvault than the one used for project destination secrets)
- [ ] If the key doesn't exist, create it and save it in the keyvault
- [ ] in memory cache the salt values (in the hashing api)
- [ ] Use this salt in secure hashing

milanmlft:
Fixes #343 

* Use `core.project_config.secrets.AzureKeyVault` in `hasher` to set up Azure Key Vault connection
* Hashing salts are now retrieved from the key vault, using the project-name as key
	* If no salt exists, a new one is created and stored in the key vault
* Removed the old `hasher` key vault functionality
* Salt generation is moved into the hashing functionality, so only `hasher` knows about it

### To do

- [x] Wait until #366 gets merged into `main` before continuing


pytest-pixl/src/pytest_pixl/plugin.py
```diff
@@ -75,3 +76,9 @@ def omop_resources() -> Path:
     traversable = resources.files("pytest_pixl").joinpath("data/omop-resources")
     with resources.as_file(traversable) as path:
         return path
+
+
+@pytest.fixture()
+def azure_keyvault() -> MockKeyVault:
```
stefpiatek, line 82: is this fixture used anywhere?

pytest-pixl/src/pytest_pixl/keyvault.py
```diff
@@ -0,0 +1,35 @@
...
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+"""A mock Azure Keyvault instance for testing."""
+
+
+class MockKeyVault:
```
stefpiatek, line 17: feels like this should just belong in hasher api tests if we're only using it there? We can then remove pytest pixl from depdendencies

pixl_core/src/core/project_config/secrets.py
```diff
@@ -80,18 +93,31 @@ def _fetch_secret(secret_name: str) -> str:
...
     secret = client.get_secret(secret_name).value
 
     if secret is None:
         msg = f"Azure Key Vault secret {secret_name} is None"
         raise ValueError(msg)
     return str(secret)
+
+
+@lru_cache
+def _create_secret(client: SecretClient, secret_name: str, secret_value: str) -> None:
```
stefpiatek, line 109: all arguments need to be hashable for caching to be enabled. So you could make SecretClient hashable or keep this as before where it was being cached

pixl_core/src/core/project_config/secrets.py
```diff
@@ -71,7 +84,7 @@ def _check_system_envvar(var_name: str) -> None:
 
 
 @lru_cache
-def _fetch_secret(secret_name: str) -> str:
+def _fetch_secret(client: SecretClient, secret_name: str) -> str:
```
stefpiatek, line 87: see below for all arguments having to be hashable. check locally that the caching is truly being used

hasher/src/hasher/endpoints.py
```diff
@@ -33,13 +33,15 @@ async def heart_beat() -> str:
     summary="Produce secure hash with optional max output length (2 <= length <= 64)",
 )
 async def hash(message: str, length: int = 64) -> Response:  # noqa: A001
-    output = generate_hash(message, length)
+    hasher = Hasher()
```
stefpiatek, line 36: could we just have a `hasher` object created upon startup of the app and we use that in all of the subcalls?

.github/workflows/main.yml
```diff
@@ -56,8 +56,11 @@ jobs:
           python-version: "3.11"
           cache: "pip"
 
-      - name: Install package
+      - name: Install Python dependencies
         run: |
+          python -m pip install --upgrade pip
+          pip install -e pixl_core
+          pip install -e pytest-pixl
```
stefpiatek, line 63: Probabyl could strip out pytest-pixl?

hasher/src/hasher/hashing.py
```diff
@@ -24,74 +24,80 @@
...
+class Hasher:
     """
-    Generate a keyed hash digest from the message using Blake2b algorithm.
-    The Azure DICOM service requires identifiers to be less than 64 characters.
-
-    :param message: string to hash
-    :param length: maximum number of characters in the output (2 <= length <= 64)
-    :return: hashed string
+    Hasher class to generate keyed hash digests using the Blake2b algorithm and fetch the
+    salt for a specific project from the Azure Key Vault instance.
```
stefpiatek, line 39:
```suggestion
    Hasher class to generate keyed hash digests using the Blake2b algorith. 
    Gets or creates the salt for a specific project from the Azure Key Vault instance.
```

hasher/src/hasher/hashing.py
```diff
@@ -24,74 +24,80 @@
...
+
+        # HMAC digest is returned as hex encoded i.e. 2 characters per byte
+        output_bytes = length // 2
+
+        key = self.keyvault.fetch_secret(config("AZURE_KEY_VAULT_SECRET_NAME"))
+        return blake2b(
+            message.encode("UTF-8"), digest_size=output_bytes, key=key.encode("UTF-8")
+        ).hexdigest()
+
+    def fetch_salt(self, project_slug: str, max_length: int = 16) -> str:
```
stefpiatek, line 76: maybe `get_or_create_salt`?

stefpiatek (commented):
Looking nice, but refactoring of caching won't actually cache sadly

hasher/src/hasher/endpoints.py
milanmlft: Yeah so I had that originally, but that's where I ran into the `pytest` setup problems, because it would create that `hasher` object during one of the imports, before my monkeypatch to use a mock key vault was applied, so it would try to set up an actual key vault connection with `AzureKeyVault()`, which would then fail because the environment variables are not set... I got really confused as to why the monkeypatch wasn't working as I expected, so ended up just moving these calls inside API calls 🤷‍♂️ 

But agreed that it's not very nice. On the other hand, I'm now actually in the process of removing all the other endpoints except for `/hash`, and moving the special handling logic for patient ID and accession number into `dcmd`, which feels more reasonable?

I'll have another look to see if I can fix the hasher object creation in a way that is compatible with the tests (or vice versa).

pytest-pixl/src/pytest_pixl/plugin.py
milanmlft: Ah yeah good point, this is leftover from an earlier attempt

pytest-pixl/src/pytest_pixl/keyvault.py
milanmlft: Yeah was originally thinking of also using it for the `core.uploader` tests, which is why I put it here, but maybe not that useful.

hasher/src/hasher/endpoints.py
stefpiatek: what are we importing in the `conftest` which ends up importing this module? Should be able to monkeypatch if we're not importing it, and I wouldn't expect the conftest to import much, but could always import locally in a fixture to avoid this?

milanmlft: I updated the `Hasher` class slightly. Now it takes the project slug as a parameter during initialisation, so now it has to be created within the API call anyway, I think?

milanmlft: Resolving for now, but open to more feedback!

milanmlft:
`pixl_dcmd` tests failing but will wait until #366 gets merged in before resolving.

milanmlft:
Docs could definitely still use some further improvement.

stefpiatek (approved):
Nice! I think some env changes are needed and a couple of questions about the public API

hasher/pyproject.toml
```diff
@@ -1,35 +1,26 @@
...
 requires-python = ">=3.10"
-classifiers = [
-    "Programming Language :: Python :: 3"
-]
+classifiers = ["Programming Language :: Python :: 3"]
 dependencies = [
+    "core",
     "azure-identity==1.12.0",
     "azure-keyvault==4.2.0",
+    "environs==9.5.0",
```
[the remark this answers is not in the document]
milanmlft: Aah this is a legacy dependency, not used anymore since I deleted `hasher.settings`. Will remove!

hasher/src/hasher/endpoints.py
```diff
@@ -32,29 +32,16 @@ async def heart_beat() -> str:
...
 )
-async def hash(message: str, length: int = 64) -> Response:  # noqa: A001
-    output = generate_hash(message, length)
+async def hash(  # noqa: A001
+    project_slug: str,
+    message: str,
+    length: int = 64,
+    *,
+    salt_length: int = 16,
+    override_salt: bool = False,
```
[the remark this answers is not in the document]
milanmlft: Yeah good point, this was already feeling a bit awkward. So would we then still want a separate `/salt` endpoint for that?

hasher/src/hasher/endpoints.py
```diff
@@ -32,29 +32,16 @@ async def heart_beat() -> str:
...
     summary="Produce secure hash with optional max output length (2 <= length <= 64)",
 )
-async def hash(message: str, length: int = 64) -> Response:  # noqa: A001
-    output = generate_hash(message, length)
+async def hash(  # noqa: A001
+    project_slug: str,
+    message: str,
+    length: int = 64,
+    *,
+    salt_length: int = 16,
```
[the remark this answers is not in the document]
milanmlft: Not sure, this was mainly taken over from the previous implementation. Though I was wondering whether this was actually ever used.

pixl_core/src/core/project_config/secrets.py
```diff
@@ -80,18 +95,38 @@ def _fetch_secret(secret_name: str) -> str:
...
+    client = _setup_client()
     secret = client.get_secret(secret_name).value
 
     if secret is None:
         msg = f"Azure Key Vault secret {secret_name} is None"
         raise ValueError(msg)
     return str(secret)
+
+
+@lru_cache
```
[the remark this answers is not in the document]
milanmlft: Ah yes, probably doesn't make sense to cache then.

milanmlft:
Had to create a new secret in our dev Azure Keyvault under `hasher--key` to get system test passing. The Hasher API reads the secret configured with `HASHER_API_AZ_KEY_VAULT_SECRET_NAME` from the vault as key for the hashing. Added this to `test/.env` as well.

PR #373, opened 2024-04-19
Delete partially transferred studies on failure

stefpiatek:
For larger studies with multiple series we get network errors fairly frequently which means that a study fails after receiving series. To ensure that a partially fulfilled study doesn't become stable and get transferred to anon, we should delete these partially transferred studies upon job failure or timeout

### DIFF
--- a/pixl_imaging/src/pixl_imaging/_processing.py
+++ b/pixl_imaging/src/pixl_imaging/_processing.py
@@ -38,10 +38,26 @@ async def process_message(message: Message) -> None:
 
     study = ImagingStudy.from_message(message)
     orthanc_raw = PIXLRawOrthanc()
+    try:
+        await _process_message(study, orthanc_raw)
+    except PixlDiscardError:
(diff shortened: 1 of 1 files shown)


milanmlft (approved):
🙌

pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -38,10 +38,26 @@ async def process_message(message: Message) -> None:
...
     orthanc_raw = PIXLRawOrthanc()
-
+    try:
+        await _process_message(study, orthanc_raw)
+    except PixlDiscardError:
+        # if a message has failed mid-transfer, can have partial study in orthanc-raw
+        # delete this so that it doesn't become stable and is exported
+        studies_to_delete = await orthanc_raw.query_local(study.orthanc_query_dict)
+        for study_to_delete in studies_to_delete:
+            logger.debug(
```
peshence, line 48: I would think this warrants at least an info level

PR #374, opened 2024-04-19
Update logging

stefpiatek:
* Use loguru throughout (haven't touched ehr api as we're removing it)
* Use loguru templating
* Changes in logging level, on feedback from @peshence


milanmlft (approved):
👍

PR #379, opened 2024-04-23
Add DICOMweb uploader

### LINKED ISSUE #375: Add DICOMWeb uploader

- [x] PIXL is able to export via dicomweb to MS PINPOINT server using basic auth (username and password)
- [x] `"dicomweb"` is a valid destination option in the project config
	- [x] Update `template_config.yaml` to reflect this

### Testing

- [x] Add unit tests for uploading to a `DICOMweb` server (probably will have to mock this out)
- [x] Add another `orthanc` container that acts as a `DICOMweb` endpoint for system tests 
- [x] Use second project in system test that is configured to use `DICOMweb`

### Documentation

- [x] Add `docs/services/dicomweb-server.md` to describe the configuration of the `DICOMweb server in system tests and how we interact with it
- [ ] Update `pixl_core/README.md`, be consistent with `## Uploading to an FTPS server`
- [x] Add relevant docstrings

### Dependencies

_No response_

### Details and Comments

Part of https://github.com/UCLH-Foundry/the-rolling-skeleton/issues/221

Should be able to recycle some functionality from the old [`SendViaStow()`](https://github.com/UCLH-Foundry/PIXL/blob/6b043f68a4b2d9480a8c3d7da6854dbb0447980e/orthanc/orthanc-anon/plugin/pixl.py#L178) in `orthanc/orthanc-anon/plugin/pixl.py`.
https://github.com/UCLH-Foundry/PIXL/blob/6b043f68a4b2d9480a8c3d7da6854dbb0447980e/orthanc/orthanc-anon/plugin/pixl.py#L178-L204

milanmlft:
This adds the `DicomWebUploader` class to allow PIXL to take `dicomweb` as an alternative endpoint to upload DICOM files to.

Fixes #375 

### Dependencies

- #380 

The new system test runs an extra project configured with `dicomweb` as its destination, but using the same input data as the original test project otherwise. This results in some ambiguous matches when querying the PIXL database as we're only querying on MRN and accession number. #380 should resolve that by including the project slug in the query.


jeremyestein:
It might help to know about this new export-api (ehr-api) endpoint in my branch:
https://github.com/UCLH-Foundry/PIXL/blob/347f25cf0ed463b0f4f1225b827084afc28f21e6/pixl_export/src/pixl_export/main.py#L90-L110
The orthanc-anon plugin now triggers the upload by calling this, rather than doing it itself. But I think as long as the API implementation gets the right uploader (which is what you'll be implementing for the DICOM upload route IIUC), then it'll be fine.

milanmlft:
Making ready to review to run system tests (which will probably break)

test/resources/omop-dicomweb/extract_summary.json
```diff
@@ -0,0 +1,94 @@
+{
```
stefpiatek, line 1: its a shame we have to duplicate this, but I think we probably do have to so that we have another endpoint for the config. Could we just slim down the json files to what we actually use?

test/docker-compose.yml
```diff
@@ -80,3 +76,29 @@ services:
...
+        image: orthancteam/orthanc:24.4.0
+        platform: linux/amd64
+        environment:
+            ORTHANC_NAME: "dicomweb"
+            ORTHANC_USERNAME: "orthanc"
+            ORTHANC_PASSWORD: "orthanc"
+            ORTHANC_AE_TITLE: "DICOMWEB"
+            RAW_AE_TITLE: ORHTANCRAW
+            RAW_DICOM_PORT: "4242"
+            RAW_IP_ADDR: "orthanc-raw" # aka. hostname
```
stefpiatek, line 90: should this be `dicomweb-server`?

test/dicomweb_config/dicom.json
```diff
@@ -0,0 +1,21 @@
...
+            "Port": "${RAW_DICOM_PORT}",
+            "Host": "${RAW_IP_ADDR}",
+            "Manufacturer": "Generic",
+            "AllowEcho": true,
+            "AllowFind": true,
+            "AllowGet": true,
+            "AllowMove": true,
+            "AllowStore": true
+        }
+    },
```
stefpiatek, line 16: should this know about raw?

test/conftest.py
```diff
@@ -136,8 +137,8 @@ def _upload_dicom_instance(dicom_dir: Path, **kwargs: Any) -> None:
 @pytest.fixture(scope="session")
 def _setup_pixl_cli(ftps_server: PixlFTPServer, _populate_vna: None) -> Generator:
     """Run pixl populate/start. Cleanup intermediate export dir on exit."""
-    # CLI calls need to have CWD = test dir so they can find the pixl_config.yml file
     run_subprocess(["pixl", "populate", str(RESOURCES_OMOP_DIR.absolute())], TEST_DIR)
+    run_subprocess(["pixl", "populate", str(RESOURCES_OMOP_DICOMWEB_DIR.absolute())], TEST_DIR)
```
stefpiatek, line 141: If this works then that's all good, may need to have a wait between the two populate commands so they don't clash. Expect that you may get a condition where the PIXL project name private tag is changed before it becomes stable so the ftps export wouldn't happen perhaps?

projects/configs/test-extract-uclh-omop-cdm-dicomweb.yaml
```diff
@@ -0,0 +1,33 @@
+#  Copyright (c) 2022 University College London Hospitals NHS Foundation Trust
```
stefpiatek, line 1: probably should be 2024 I'd guess

pixl_core/tests/uploader/test_dicomweb.py
```diff
@@ -0,0 +1,73 @@
...
+    stow_response = dicomweb_uploader.send_via_stow(study_id)
+    studies_response = _do_get_request("/dicom-web/studies", data={"Uri": "/instances"})
+    servers_response = _do_get_request("/dicom-web/servers")
+
+    # ASSERT
+    # Check if dicom-web server is set up correctly
+    assert DICOM_ENDPOINT_NAME in servers_response.json()
+    assert stow_response.status_code == 200  # succesful upload
+    # Taken from https://orthanc.uclouvain.be/hg/orthanc-dicomweb/file/default/Resources/Samples/Python/SendStow.py
+    assert "00081190" in studies_response.json()[0]
```
stefpiatek, line 73: Could you add a comment here to explain what this is checking

pixl_core/tests/uploader/test_dicomweb.py
```diff
@@ -0,0 +1,73 @@
...
+from core.uploader._dicomweb import DicomWebUploader  # type: ignore [import-untyped]
+from decouple import config  # type: ignore [import-untyped]
+
+ORTHANC_URL = config("ORTHANC_URL")
+DICOM_ENDPOINT_NAME = config("DICOM_ENDPOINT_NAME")
+ORTHANC_USERNAME = config("ORTHANC_USERNAME")
+ORTHANC_PASSWORD = config("ORTHANC_PASSWORD")
+
+
+class MockDicomWebUploader(DicomWebUploader):
```
stefpiatek, line 31: tbf as this has some decent testing of dicom web here, could just not have the system test version as we know that the overall uploader flow works

pixl_core/src/core/uploader/_ftps.py
```diff
@@ -12,7 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-"""Helper functions for setting up a connection to an FTPS server."""
+"""Uploader subclass for FTPS uploads."""
```
stefpiatek, line 15:
```suggestion
"""Uploader subclass for FTPS."""
```

docs/services/dicomweb-server.md
```diff
@@ -0,0 +1,24 @@
+# DICOMweb server
```
stefpiatek, line 1: Is this a service? I would had thought it should be documented as an endpoint as we only have this during testing

pixl_core/src/core/uploader/_dicomweb.py
```diff
@@ -0,0 +1,121 @@
...
+        super().__init__(project_slug, keyvault_alias)
+
+    def _set_config(self) -> None:
+        # Use the Azure KV alias as prefix if it exists, otherwise use the project name
+        az_prefix = self.keyvault_alias
+        az_prefix = az_prefix if az_prefix else self.project_slug
+
+        self.user = self.keyvault.fetch_secret(f"{az_prefix}--dicomweb--username")
+        self.password = self.keyvault.fetch_secret(f"{az_prefix}--dicomweb--password")
+        self.orthanc_url = config("ORTHANC_URL")
```
stefpiatek, line 43: Should this be orthanc anon?
```suggestion
        self.orthanc_url = config("ORTHANC_ANON_URL")
```

pixl_core/src/core/uploader/_dicomweb.py
```diff
@@ -0,0 +1,121 @@
...
+                data=json.dumps(payload),
+                timeout=30,
+            )
+        except requests.exceptions.RequestException:
+            logger.error("Failed to send via stow")
+            raise
+        else:
+            if response.status_code != success_code:
+                logger.error("Upload via stow unsuccesful")
+                raise RuntimeError
```
stefpiatek, line 75: is there a reason we're not using raise_for_status here?

stefpiatek (approved):
Looking nice, thanks Milan. May not even need to have the system test for dicomweb but not too strongly held on that

test/resources/omop-dicomweb/extract_summary.json
jeremyestein: We could look into some sort of "smart" json solution such as jsonnet so we don't just have to copy and paste everything.

milanmlft: >its a shame we have to duplicate this, but I think we probably do have to so that we have another endpoint for the config. Could we just slim down the json files to what we actually use?
I agree but this would require some knowledge of what all these fields actually mean and what they're used for... which I currently don't have 😅

pixl_core/tests/uploader/test_dicomweb.py
milanmlft: Hmm yeah, I actually thought of removing the unit test from core and keep the system test, as the latter is a bit more realistic, but I'm fine with either :)

docs/services/dicomweb-server.md
milanmlft: Ah yeah good point, I might just move this to `pixl_core/README`

test/docker-compose.yml
stefpiatek: then shall we remove it?

pixl_core/tests/uploader/test_dicomweb.py
stefpiatek: mm yeah, let's leave both in for now and we can always reassess how this affects development later on maybe?

test/resources/omop-dicomweb/extract_summary.json
stefpiatek: we basically only care about the project name and the datetime. Look through the code for what we use and strip out everything else?

test/docker-compose.yml
stefpiatek: oh actually its the `RAW_IP_ADDR` ignore me

pixl_core/src/core/uploader/_dicomweb.py
milanmlft: So as this could is typically run from within `orthanc-anon`, this should point to `http://localhost:8042`, which is the default Orthanc URL as I understand it and which is why I give it a more generic name I think.

pixl_core/src/core/uploader/_dicomweb.py
milanmlft: Errr nope! Will change :)

pixl_core/src/core/uploader/_dicomweb.py
stefpiatek: Fair enough, will then want to update it when Jeremy pulls this into the export api

PR #380, opened 2024-04-25
Include `project_slug` in Image PIXL database queries

milanmlft:
This should allow running multiple projects with the same or overlapping studies without having to
tear down and rebuild the entire setup.

Necessary for #379.

### DIFF
--- a/pixl_dcmd/src/pixl_dcmd/_database.py
+++ b/pixl_dcmd/src/pixl_dcmd/_database.py
@@ -14,9 +14,9 @@
 
 """Interaction with the PIXL database."""
 
-from decouple import config
+from decouple import config  # type: ignore [import-untyped]
 
-from core.db.models import Image
+from core.db.models import Image, Extract  # type: ignore [import-untyped]
 from sqlalchemy import URL, create_engine
 from sqlalchemy.orm import sessionmaker
 
@@ -53,12 +53,14 @@ def add_hashed_identifier_and_save_to_db(
         return updated_image
 
 
-def query_db(mrn: str, accession_number: str) -> Image:
+def query_db(project_slug: str, mrn: str, accession_number: str) -> Image:
     PixlSession = sessionmaker(engine)
     with PixlSession() as pixl_session, pixl_session.begin():
         existing_image: Image = (
             pixl_session.query(Image)
+            .join(Extract)
             .filter(
+                Extract.slug == project_slug,
                 Image.accession_number == accession_number,
                 Image.mrn == mrn,
                 Image.exported_at == None,  # noqa: E711
--- a/pixl_dcmd/src/pixl_dcmd/_database.py
+++ b/pixl_dcmd/src/pixl_dcmd/_database.py
@@ -16,7 +16,7 @@
 
 from decouple import config  # type: ignore [import-untyped]
 
-from core.db.models import Image, Extract  # type: ignore [import-untyped]
+from core.db.models import Image, Extract
 from sqlalchemy import URL, create_engine
 from sqlalchemy.orm import sessionmaker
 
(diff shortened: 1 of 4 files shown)


cli/src/pixl_cli/_database.py
```diff
@@ -14,8 +14,9 @@
 
 """Interaction with the PIXL database."""
 
-from core.db.models import Extract, Image
-from core.patient_queue.message import Message
+from core.db.models import Extract, Image  # type: ignore [import-untyped]
+from core.patient_queue.message import Message  # type: ignore [import-untyped]
+from slugify import slugify  # type: ignore [import-untyped]
```
stefpiatek, line 19: how come these are popping up just now?

pixl_core/src/core/db/models.py
```diff
@@ -48,6 +48,7 @@ class Image(Base):
 
     __tablename__ = "image"
 
+    project_name: Mapped[str]
```
stefpiatek, line 51: oh this is already in the `Extract` table

cli/src/pixl_cli/_database.py
```diff
@@ -85,6 +90,7 @@ def _get_image_and_check_exported(
         session.query(Image)
         .filter(
             Image.extract == extract,
+            Image.project_name == slugify(message.project_name),
```
stefpiatek, line 93: oh this is already in the `Extract` table, so will need to query through that, instead of adding it to the Image

pixl_dcmd/src/pixl_dcmd/_database.py
```diff
@@ -53,12 +54,13 @@ def add_hashed_identifier_and_save_to_db(
...
 
 
-def query_db(mrn: str, accession_number: str) -> Image:
+def query_db(project_slug: str, mrn: str, accession_number: str) -> Image:
     PixlSession = sessionmaker(engine)
     with PixlSession() as pixl_session, pixl_session.begin():
         existing_image: Image = (
             pixl_session.query(Image)
             .filter(
+                Image.project_name == project_slug,
```
stefpiatek, line 63: then this would also query through the `Extract` table to make sure its the right project

stefpiatek (changes requested):
Does work, but also the `Extract` table has that data so we're at least a bit more normalised

cli/src/pixl_cli/_database.py
milanmlft: 🤷‍♂️ tbf they typically pass pre-commit but keep popping up locally for me in my editor

milanmlft: `Skipping analyzing "pytest_pixl.dicom": module is installed, but missing library stubs or py.typed marker`
Might just be a configuration thing maybe

pixl_dcmd/src/pixl_dcmd/_database.py
```diff
@@ -14,9 +14,9 @@
 
 """Interaction with the PIXL database."""
 
-from decouple import config
+from decouple import config  # type: ignore [import-untyped]
 
-from core.db.models import Image
+from core.db.models import Image, Extract  # type: ignore [import-untyped]
```
stefpiatek, line 19: mm not sure why this is happening

pixl_dcmd/src/pixl_dcmd/_database.py
```diff
@@ -53,12 +53,14 @@ def add_hashed_identifier_and_save_to_db(
...
 
-def query_db(mrn: str, accession_number: str) -> Image:
+def query_db(project_slug: str, mrn: str, accession_number: str) -> Image:
     PixlSession = sessionmaker(engine)
     with PixlSession() as pixl_session, pixl_session.begin():
         existing_image: Image = (
             pixl_session.query(Image)
+            .join(Extract)
             .filter(
+                Extract.slug == project_slug,
```
stefpiatek, line 63: ah 3 lines, I was close

stefpiatek (approved):
Nice!

pixl_dcmd/src/pixl_dcmd/_database.py
milanmlft: Ah woops, thought I reverted all of them

PR #386, opened 2024-05-01
Add healthchecks to orthanc servers in imaging test

jeremyestein:
I noticed the imaging test was failing intermittently. The error occurs when trying to upload test data with curl. It wasn't waiting for orthanc to be up before running this command.

Running pytest on command line:
Without fix:
Failed 8 times out of 100

With fix:
Failed 0 times out of 100


stefpiatek (approved):
Nice catch, thanks Jeremy

Issue #239, opened 2024-01-26
Audit services exposed on GAE


jeremyestein:
### Definition of Done / Acceptance Criteria

We will have demonstrated to the best of our knowledge that the (micro-)services we are running:
* have the minimum exposure necessary to perform their function, in the sense of whether the ports are accessible to the whole hospital, just the GAE, or not at all
* have sufficient security controls (eg. authentication, source IP restriction) where some exposure is needed
* will have security updates applied quickly when needed

### Testing

Could imagine some tests being made automatically as part of a system test. Eg. run a `netstat -an` before and after or something?
Some might have to be manual.
Some of this is about raising awareness that will feed into code reviews, etc.

### Documentation

This should definitely be documented for the sake of future developers. This ties into our action coming away from the TI planning retrospective or having better low-level/design documentation.

### Details and Comments

This would mainly involve going through our docker-compose files, bringing up a test system on the GAE and seeing what ports it opens, listing the authentication/etc status of each service in a short document.
The dev team can do this ourselves, but we may also need external expert input to make sure we haven't overlooked something.

stefpiatek:
Suggest starting off doing an audit which lists what containers, what ports are exposed. Then development team would use this to determine whether this is correct

dram1964:
Looking into [docker scout](https://docs.docker.com/scout/)

stefpiatek:
scout: difficult to scan containers instead of images. Will develop best practices for running containers

dram1964:
Partly covered by SOPs, but need to start building some actual monitoring processes on the GAEs for:
- [ ] Ports exposed, network visibility
- [ ] CVEs
- [ ] authentication settings for services

Additional to collecting this data, we would also need to consider what resources are available to monitor the data collected. Perhaps some dashboards would ease the burden

dram1964:
Now collecting audit data for:
1. Ports exposed externally on the GAEs for each running container
2. UID:GID for each running container

dram1964:
CVE monitoring needs to be rolled out to each GAE:
- [ ] GAE01
- [ ] GAE02
- [ ] GAE03
- [ ] GAE04
- [X] GAE05
- [ ] GAE06
- [ ] GAE07
- [ ] GAE08
- [ ] GAE09
- [ ] GAE10
- [ ] GAE11
- [ ] GAE12

PR #387, opened 2024-05-01
Don't expose ports on all interfaces unless needed

jeremyestein:
Partially addresses #239 (fixing rather than auditing)

Most open ports only need to be accessed from the CLI that's running on the same host. So only listen on localhost.

Similar principal applies to running tests, although those usually won't be run in the hospital.

Orthanc raw and anon are a notable exception as someone may want to view those from the DSD or another hospital computer, but otherwise exposing everything to (perhaps) the whole hospital is a security hazard.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -244,7 +244,7 @@ services:
             retries: 3
         ports:
             - "127.0.0.1:${RABBITMQ_PORT}:5672"
-            - "127.0.0.1:${RABBITMQ_ADMIN_PORT}:15672"
+            - "${RABBITMQ_ADMIN_PORT}:15672"
         networks:
             - pixl-net
         volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -94,7 +94,7 @@ services:
         env_file:
             - ./docker/common.env
         ports:
-            - "${HASHER_API_PORT}:8000"
+            - "127.0.0.1:${HASHER_API_PORT}:8000"
         volumes:
             - *logs-volume
         networks:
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -243,8 +243,8 @@ services:
             timeout: 30s
             retries: 3
         ports:
-            - "${RABBITMQ_PORT}:5672"
-            - "${RABBITMQ_ADMIN_PORT}:15672"
+            - "127.0.0.1:${RABBITMQ_PORT}:5672"
+            - "127.0.0.1:${RABBITMQ_ADMIN_PORT}:15672"
```
stefpiatek, line 247: we want the admin port to be exposed so that we can monitor the status of the queue using the admin page 
```suggestion
            - "${RABBITMQ_ADMIN_PORT}:15672"
```

docker-compose.yml
```diff
@@ -354,7 +354,7 @@ services:
               source: postgres-data
               target: /var/lib/postgresql/data
         ports:
-            - "${POSTGRES_PORT}:5432"
+            - "127.0.0.1:${POSTGRES_PORT}:5432"
```
stefpiatek, line 357: we also want this to be exposed to the network so we can monitor the progress from DSD

stefpiatek (approved):
Nice Jeremy, a couple of them we do need to be accessible outside of the GAE but happy for you to merge when you've resolved

docker-compose.yml
jeremyestein: good point!

PR #389, opened 2024-05-02
feat: replace hashed_id with pseudo study uid

peshence:
Replace linking variable

from: PatientID=hash(PatientID+AccesseionNumber)
to: StudyInstanceUID=generate_uid()

Database is queried to check if a UID was already assigned to this image (identified by patient id and accession number for now, but to be switched to StudyUID once OMOP ES starts serving that)


pixl_core/src/core/db/models.py
```diff
@@ -52,7 +52,7 @@ class Image(Base):
     accession_number: Mapped[str]
     study_date: Mapped[Date] = mapped_column(Date())
     mrn: Mapped[str]
-    hashed_identifier: Mapped[Optional[str]]
+    pseudo_study_uid: Mapped[Optional[str]]
```
stefpiatek, line 55: You'll also need to create an alembic migration for this to be able to be deployed in production

peshence: I did? https://github.com/UCLH-Foundry/PIXL/pull/389/files#diff-bf17b0fcdc74b312ccafae120d2152b11b53489f1380d128086ee707e9f81a9e

stefpiatek: ha oh sorry, I swore I looked for them

stefpiatek (approved):
Ah great thanks for getting this in. Couple of small changes. I'd really like to the pseudo image identifiers that have already been run in production, imagine we can rename, if not then we can just have the column there but its never used.

test/system_test.py
```diff
@@ -175,7 +179,20 @@ def _check_dcm_tags_from_zip(
...
+                    "instances"
+                ]
+            )
+        elif len(actual_instances) == 1:
+            assert (
+                actual_instances
+                == expected_studies["1.3.46.670589.11.38023.5.0.14068.2023012517090160002"][
+                    "instances"
+                ]
+            )
```
jeremyestein, line 195: Why is the number of expected instances variable? I'm assuming this is something to do with the unpredictability of the UID, but then why is there a UID specified here at all if the actual value will never match it? Could we maybe use the procedure_occurrence_id as a fixed ID to identify the right one to compare against?

pixl_dcmd/src/pixl_dcmd/_database.py
```diff
@@ -33,40 +34,82 @@
...
+
+    (
+        pixl_session.query(Image)
+        .filter(
+            Image.accession_number == existing_image.accession_number,
+            Image.mrn == existing_image.mrn,
+            Image.pseudo_study_uid == pseudo_study_uid,
         )
+        .one()
+    )
```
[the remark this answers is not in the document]
peshence: I think I left it as "not sure what this is for, I'll check later" and forgot, good spot

PR #400, opened 2024-05-08
Don't run docker containers as root

### LINKED ISSUE #234: Do not run containers as root and do not create root owned files

jeremyestein:
Fixes #234. Waiting for testing on GAE before merging.

Note the addition of two new variables `PIXL_USER_UID` and `PIXL_USER_GID`

Firstly merge all the Dockerfiles for images that we control (imaging, export, hasher) to make this process easier.

Run all our python containers as the user/group pixl, which we create as part of the build process, using the UID/GID as specified in the config.

Export API mounts export dir read-only as it doesn't need to write any more.

Document how the host must be set up for this to work.

Do same for orthanc images.


jeremyestein:
In writing the docs, I have come up with what I think is a problem, that will likely need a user/group adding to the GAE to fix. Let me quote from the docs:

---

Imagine you are a user on the GAE with username `fred`, primary group `fred`, and supplementary
group `docker`.

When you run the CLI, it will run as `fred:fred`, and normally any files it creates would also have
this ownership. (ACLs defined in `/gae` will affect this however, so the files may have
a different group ownership, such as `fred:docker`)

There is a problem here in that the export API will likely be running as some other user/group,
so it will not be able to read the files.

Options for fixing:
* The CLI (or an ACL on the exports dir) makes the files world readable. Not ideal from a security
  pov.
* Relying on the existing ACL for `/gae` which gives files group ownership of `docker`, we run the
   Export API as the `docker` group, thus giving it read access to these files. However, this is
   very similar to running the container as root, which is what this whole thing is trying to avoid.
* Create a new group (`pixl`) on the GAE and add all pixl devs to it.
  Set an ACL on the exports dir that sets group ownership for all created files as `pixl`.
  Then we could set its GID in the env file so that the container runs as group `pixl`.
* Always build and run the containers as the same user that is running the CLI. You set the
  build vars `PIXL_USER_[UG]ID` to the numerical value of fred:fred. PIXL runs as you. When you run
  CLI, the files are owned by fred:docker, so export api can read them fine. This lacks flexibility as we should
  all be able to run PIXL.

docs/design/docker_perms.md
```diff
@@ -0,0 +1,83 @@
...
+  pov.
+* Relying on the existing ACL for `/gae` which gives files group ownership of `docker`, we run the
+   Export API as the `docker` group, thus giving it read access to these files. However, this is
+   very similar to running the container as root, which is what this whole thing is trying to avoid.
+* Create a new group (`pixl`) on the GAE and add all pixl devs to it.
+  Set an ACL on the exports dir that sets group ownership for all created files as `pixl`.
+  Then we could set its GID in the env file so that the container runs as group `pixl`.
+* Always build and run the containers as the same user that is running the CLI. You set the
+  build vars `PIXL_USER_[UG]ID` to the numerical value of fred:fred. PIXL runs as you. When you run
+  CLI, the files are owned by fred:docker, so export api can read them fine.
```
peshence, line 70: I also just wrote this on Slack, but for posterity:

> I assume that we intend to have the apis running in the background indefinitely, started by a specific role user and that multiple different users may start the pipeline from the CLI, so the new group approach makes the most sense to me.
> I also like the idea of the CLI not actually containing functionality and only accessing APIs, but would say that is a separate issue/PR

This portion of the docs would then be converted to just instructions about having the group, do I understand correctly?

test/conftest.py
```diff
@@ -135,43 +134,19 @@ def _upload_dicom_instance(dicom_dir: Path, **kwargs: Any) -> None:
...
-        [
-            "docker",
-            "exec",
-            "system-test-export-api-1",
-            "rm",
-            "-r",
-            "/run/projects/exports/test-extract-uclh-omop-cdm/",
-        ],
-        TEST_DIR,
-    )
```
peshence, line 154: 🚀

.github/workflows/main.yml
```diff
@@ -128,6 +119,18 @@ jobs:
                   HASHER_API_AZ_TENANT_ID: ${{ secrets.EXPORT_AZ_TENANT_ID }}
                   HASHER_API_AZ_KEY_VAULT_NAME: ${{ secrets.EXPORT_AZ_KEY_VAULT_NAME }}
               run: |
+                  # We need PIXL_USER_UID and PIXL_USER_GID from the test env file.
+                  source .env
+                  # Export API (running as pixl) needs to be able to *read* files in
+                  # the export dir, but it doesn't write there.
+                  # So, set ownership/permissions accordingly.
+                  # Bear in mind that ownership/perms of files created below this dir will
+                  # depend on the CLI process. I think we don't need to set ACLs for now.
```
peshence, line 128:
```suggestion
                  # Export API (running as pixl) needs only *read* files in the export dir.
                  # Ownership/perms of files created in the export dir will
                  # depend on the user running the CLI process.
```

Not adamant about the wording but would remove the opinions/questions and make it shorter and more to the point

peshence (approved):
Looks good

docs/design/docker_perms.md
jeremyestein: Yes, it would be converted to instructions as soon as I work out what those instructions are. Then this bit could perhaps go into the body of an issue that recommends refactoring the CLI.

jeremyestein:
This change can't go in until the `pixl` user and group have been created on the GAE, and the pixl dev users added to the group.

dram1964:
> This change can't go in until the `pixl` user and group have been created on the GAE, and the pixl dev users added to the group.

Hi @jeremyestein, I've added PIXL user and group to GAE05 - and added you to the group. Let me know how that works for you.

dram1964 (approved):
Using a 'service account' for this sort of thing seems to be a good idea

README.md
```diff
@@ -295,10 +295,14 @@ These files will subsequently be uploaded to the `parquet` destination specified
 [project config](#3-configure-a-new-project).
 
 ```
-EXPORT_ROOT/PROJECT_SLUG/all_extracts/EXTRACT_DATETIME/radiology/radiology.parquet
+EXPORT_ROOT/PROJECT_SLUG/all_extracts/EXTRACT_DATETIME/radiology/IMAGE_LINKER.parquet
 ....................................................../omop/public/*.parquet
 ```
 
+This directory is current located in `projects/exports` relative to the source code root.
```
milanmlft, line 302:
```suggestion
This directory is currently located in `projects/exports` relative to the source code root.
```

docker/hasher-api/Dockerfile
milanmlft, line 1: ❤️

milanmlft (approved):

docker/orthanc/Dockerfile
```diff
@@ -37,17 +40,48 @@ COPY ./pixl_dcmd/ /pixl_dcmd
...
+
+
+# Orthanc-anon
+# =================================================================================================
+FROM pixl_orthanc_base AS pixl_orthanc_anon
+
+COPY --chown=orthanc:orthanc ./orthanc/orthanc-anon/plugin/pixl.py /etc/orthanc/pixl.py
+COPY --chown=orthanc:orthanc ./orthanc/orthanc-anon/config /run/secrets
+
 RUN sed -i "s/\${ORTHANC_CONCURRENT_JOBS}/${ORTHANC_CONCURRENT_JOBS:-5}/g" /run/secrets/orthanc.json
```
jeremyestein, line 85: `ORTHANC_CONCURRENT_JOBS` doesn't seem to be declared as a build ARG for `pixl_orthanc_anon`. Is it just reverting to the given default value?

jeremyestein:
Have split the refactoring side of this into #583 , which could possibly be merged with less risk than this PR. If we do decide to merge this, I think the fixup commit in the other branch (98fa6896fb33bf6090fb5b9c2309dbe64ce331bd) should be applied here too.

stefpiatek:
I expect this is failing because the `PIXL_DB_*` env variables were removed from anon?

stefpiatek:
Closing as has gone stale

PR #403, opened 2024-05-16
feat[cli]: only export linker information for processed images

peshence:
attempt at solving the second definition of done cirteria, closes https://github.com/UCLH-Foundry/the-rolling-skeleton/issues/250

### DIFF
--- a/cli/src/pixl_cli/_database.py
+++ b/cli/src/pixl_cli/_database.py
@@ -123,6 +123,6 @@ def processed_images_for_project(project_slug: str) -> list[Image]:
             session.query(Image)
             .join(Extract)
             .filter(Extract.slug == project_slug)
-            .filter(not_(Image.pseudo_study_uid.is_(None)))
+            .filter(not_(Image.exported_at.is_(None)))
             .all(),
         )
--- a/cli/src/pixl_cli/_database.py
+++ b/cli/src/pixl_cli/_database.py
@@ -18,7 +18,7 @@ from typing import cast
 
 from core.db.models import Extract, Image
 from core.patient_queue.message import Message
-from sqlalchemy import URL, create_engine
+from sqlalchemy import URL, create_engine, not_
 from sqlalchemy.orm import Session, sessionmaker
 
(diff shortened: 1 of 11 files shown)


cli/src/pixl_cli/_database.py
```diff
@@ -114,11 +114,15 @@ def _add_new_image_to_session(extract: Extract, message: Message, session: Sessi
...
     """Given a project, get all images in the DB for that project."""
     PixlSession = sessionmaker(engine)
     with PixlSession() as session:
         return cast(
             list[Image],
-            session.query(Image).join(Extract).filter(Extract.slug == project_slug).all(),
+            session.query(Image)
+            .join(Extract)
+            .filter(Extract.slug == project_slug)
+            .filter(not_(Image.pseudo_study_uid.is_(None)))
```
stefpiatek, line 126: ooh avoiding the `== None` is fun. Might be worth making a separate PR where you find any time we've done that and replace with nicer syntax

cli/src/pixl_cli/_database.py
```diff
@@ -114,11 +114,15 @@ def _add_new_image_to_session(extract: Extract, message: Message, session: Sessi
     return new_image
 
 
-def images_for_project(project_slug: str) -> list[Image]:
+def processed_images_for_project(project_slug: str) -> list[Image]:
```
stefpiatek, line 117: would be good to have a test which ensures that this works as we'd expect (and stops us reverting this behaviour)

cli/src/pixl_cli/_database.py
stefpiatek: Also think that maybe this should be where `exported_at` isn't empty instead of `pseudo_study_uid`. Feels like you want to know when something has been exported and that you should expect to find it

stefpiatek (approved):
Looks good, thanks for finishing this off 🎉 

Some small suggestions

cli/src/pixl_cli/_database.py
peshence: oof good catch

PR #406, opened 2024-05-29
Add in extra private tags for philips

stefpiatek:
Feedback from Prostate MRI is that some public tags can have rounding issues, so including private tags. Not sure if its worth just creating a separate override for mri's as a whole?

closes https://github.com/UCLH-Foundry/the-rolling-skeleton/issues/245


milanmlft (approved):
Looks good to me, though I can't speak to the `may need to have this for all MRs rather than just DW?` comments.

HChughtai (commented):
Sorry, I missed this PR! These parameters are applicable to all MR sequences - so if there are issues with the public tags for this manufacturer/model, we'd likely want them for all MRs rather than just DWI

milanmlft (approved):
👌

PR #407, opened 2024-05-30
Retry processing of messages until no change between runs

### LINKED ISSUE #393: Allow CLI to retry processing until no new messages are exported

As someone exporting data from PIXL, I'd like to be able to set a run off that will deal with the VNA being flaky at returning imaging studies, and it will keep on retrying unexported images until a round of processing doesn't export any images

### Testing

could use this for one of the system tests to ensure that it waits until rabbitmq is empty, but with a shorter timeout for checking whether there has been a change in exported images. Other option is to mock out the interaction with database, queue and time to be able to control this precisely and have quick tests

### Documentation

Create documentation for use

### Details and Comments

Currently have to do this manually to compensate for the VNA, to serve projects easily we should automate this.

The VNA has to pull data from long term storage to a cache layer and often fails during our request for the image. If there's a retry within 24 hours, it then works because the data has been transferred to the cache. Currently the pipline is run a few times over. As we filter out exported images before publishing messages to rabbitmq, requerying is reasonable.

Basic thoughts on how this might work

- define `last_exported_count`, number of messages exported at the start of the run. This would be 0 if the batch is new, otherwise queried from the database.
- populate messages 
- poll rabbitmq until the rabbitmq imaging queue is empty
- wait 5 minutes to allow for orthanc raw and orthanc anon to process and the image to be published
- query database for the current number of messages exported for the project. If this is the same as `last_exported_count`, success and exit! If not then query again

Might be worth setting a CLI argument for maximum number of retries, maybe 5?

stefpiatek:
closes #393

- [x] Allow env configuration of wait time for queues to empty before checking? Would be good to not wait during system tests


cli/src/pixl_cli/main.py
```diff
@@ -103,11 +109,21 @@ def check_env(*, error: bool, sample_env_file: click.Path) -> None:
...
+    "--num-retries",
+    "num_retries",
+    type=int,
+    show_default=True,
+    default=5,
+    help="Number of retries to attempt before giving up, 5 minute wait inbetween",
 )
-def populate(
-    parquet_path: Path, *, queues: str, start_processing: bool, rate: Optional[float]
+def populate(  # too many args
```
milanmlft, line 120: Could we have `--num-retries` be part of `pixl start` instead? That would avoid the `too many args`.
And in any case, presumably it only make sense to retry processing if the processing is actually started? So `pixl populate --no-start ... --num-retries 5` doesn't make much sense but is a valid command currently.

milanmlft (approved):
Looks good to me, just a small suggestion/question regarding where `--num-retries` is/should be used.

cli/src/pixl_cli/main.py
stefpiatek: `pixl start` does a different thing though, it doesn't populate so you can't have the retries based off it. Could add a check that there's logical consistency between `--no-start` and `--num-retries`?

milanmlft: Aah I see, in that case maybe just a log message/warning that `--num-retries` is ignored in case of `--no-start`

stefpiatek: done!

cli/src/pixl_cli/main.py
```diff
@@ -142,6 +142,9 @@ def populate(  # too many args
     queues_to_populate = queues.split(",")
     if start_processing:
         _start_or_update_extract(queues=queues_to_populate, rate=rate)
+    else:
+        logger.info("Starting to process messages disabled, setting `--num-retries` to 0")
+        num_retries = 0
```
milanmlft, line 147: 👌

PR #411, opened 2024-06-13
Move `pixldc` commands into PIXL CLI

### LINKED ISSUE #392: Bring pixldc into pixl CLI

As someone who runs the pixl pipeline, I'd like to control docker using the CLI and to create services in the background 

- [ ] use the pixl CLI to control docker. 
- [ ] run docker in the background, instead of foreground.

### Testing

should update the system tests to use CLI

### Documentation

should update documentation referencing pixldc

### Dependencies

_No response_

### Details and Comments

- Strange that I have to use a different command. 
- We have dependency that will take down required services if they fail, so don't need to take down everything if it fails
- I think if you run things in the foreground, docker compose makes it difficult for you to take down the running containers

milanmlft:
Fixes #392 

- **Add `pixl dc` CLI command as a wrapper around `docker compose`
- **Remove `pixldc`**
- **Update docs**

Periphery changes:
- **Use common `pixl-db` envvars for `orthanc-anon`**
- **Remove obsolete `cli._utils`**


cli/src/pixl_cli/_docker_commands.py
```diff
@@ -0,0 +1,91 @@
...
+    # Construct the docker-compose arguments
+    docker_args = ["down", *extra_args]
+    run_docker_compose(project, env_file, docker_args, working_dir=PIXL_ROOT)
+
+
+def run_docker_compose(
+    project: str, env_file: Path, args: list, working_dir: Optional[Path]
+) -> None:
+    """Wrapper to run docker-compose through the CLI."""
+    docker_compose_cmd = shutil.which("docker-compose")
```
stefpiatek, line 75: could be wrong but I think `docker-compose` is v1, and we've got v2 on the GAEs which is `docker compose`. 

Also, why is the command not just `docker compose` and let the system to work out where the executable is? You'd get a similar output if its not found in the path and just adds extra steps

docker-compose.yml
```diff
@@ -115,7 +115,7 @@ services:
         platform: linux/amd64
         command: /run/secrets
         environment:
-            <<: [*proxy-common, *pixl-common-env, *azure-keyvault]
+            <<: [*pixl-db, *proxy-common, *pixl-common-env, *azure-keyvault]
```
stefpiatek, line 118: 👌🏻

stefpiatek (commented):
Looking nice, adding in a couple of suggestions, we can try them out on the GAE this afternoon as well

cli/src/pixl_cli/_docker_commands.py
milanmlft: Yeah the `shutil.which()` is because originally I used `subprocess.Popen`, [which recommends doing this](https://docs.python.org/3/library/subprocess.html#subprocess.Popen), but for `subprocess.run()` it might be fine not to do it?

cli/src/pixl_cli/_docker_commands.py
```diff
@@ -0,0 +1,91 @@
...
+import click
+from loguru import logger
+
+ALLOWED_PROJECT_NAMES = ["pixl_dev", "pixl_test", "pixl_prod"]
+PIXL_ROOT = Path(__file__).parents[3].resolve()
+
+docker_project_option = click.option(
+    "-p",
+    "--project",
+    type=click.Choice(ALLOWED_PROJECT_NAMES, case_sensitive=False),
```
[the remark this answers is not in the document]
milanmlft: Ah yes makes sense, this was mostly carried over from `pixldc`

milanmlft:
Reverted back to using `docker compose` directly in `run-system-test.sh` because attempting to use `pixl up/down` requires the following:

- They need to be run from the project root directory (though the exact reason for this is still vague to me)
- They require all of the environment variables in `test/.env` to be set, because running the CLI through Python results in a series of imports and `config()` calls, which will fail when the envvar is not set

There's a couple of workarounds for this, like temporarily copying over the `test/.env` file to project root, or loading the `test/.env` file through Python. But I'm going to leave it at this because this PR already took way too much time for what should have been a simple change 🫠 

More importantly though, I think this points to some not-so-ideal design choices in the way we're handling these environment variables and `.env` files. Because when I'm running `pixl up`, I don't really care whether `RABBITMQ_HOST` is defined or not, but at the moment that will result in an error. Might be worth having a think about!

cli/src/pixl_cli/_docker_commands.py
```diff
@@ -0,0 +1,88 @@
...
+)
+docker_extra_args = click.argument("extra_args", nargs=-1, type=click.UNPROCESSED)
+
+
+# Required to allow passing unkown options to docker-compose
+# https://click.palletsprojects.com/en/8.1.x/advanced/#forwarding-unknown-options
+@click.command(context_settings={"ignore_unknown_options": True})
+@docker_env_option
+@docker_extra_args
+def up(env_file: list[Path], *, extra_args: tuple[str]) -> None:
```
stefpiatek, line 42: I think we might want to have this instead as `pixl dc`, then we can use all of docker compose subcommands. Then if `up` is in the args, then we add in the up args, and the check for down

cli/src/pixl_cli/_docker_commands.py
```diff
@@ -0,0 +1,88 @@
...
+    docker_args = [
+        docker_cmd,
+        "compose",
+        "--project-name",
+        f"pixl_{pixl_env}",
+        # env_file will be a list of paths, so we need to flatten it
+        *[f"--env-file={f}" for f in env_file],
+        *args,
+    ]
+    logger.debug("Running docker compose with: {}, from {}", docker_args, working_dir)
```
stefpiatek, line 86: This feels like something that you'd probably always want to know, so maybe put as `INFO`

stefpiatek (approved):
⭐ All working nicely on the GAE, though couldn't do dry run with the version of docker so tested out the prod volumes locally. 

I think probably best to make just a `pixl dc` subcommand which is a wrapper around a subprocess to `docker compose`, as the aim is to make this an easier experience and then we don't have to document what the args are for docker compose etc.

Happy for you to merge once you've done that

PR #414, opened 2024-06-21
Validate anonymised Dicom tags

### LINKED ISSUE #377: Validate processed DICOM against standard

As a consumer of data, the generated DICOM should be standard compliant

- [x] DICOM that has been processed by the PIXL pipeline should be validated against the DICOM standard (e.g. using https://pydicom.github.io/dicom-validator) in production - throw error if not and don't export

### Testing

- Validate known standards-compliant DICOM and ensure it passes checks
- Remove required tags, and ensure that the non-compliance is caught
- [x] negative test case for non-compliant dicom

### Documentation

_No response_

### Dependencies

_No response_

### Details and Comments

Currently we have no check that modified DICOM (e.g. through de-identification) complies with the standard in terms of the tags included/modified. This means that de-identfication schemes and other processing has the potential to generate non-compliant DICOM. This can cause issues with downstream tools and analysis that may not accept the DICOM, or process it incorrectly due to missing tags (e.g. if some needed tags are missing from the MRI image module, then conversion to Nifti using dcm2niix can be in the wrong orientation).

milanmlft:
- **Remove obsolete `pixl_dcmd/bin/run-tests.sh` script**
- **Implement dicom tag validation**
- **Validate Dicom after anonymisation and before export**
- **Let validation return `True` if successful**
- **Include Dicom validation in unit tests**

Fixes #377

## Notes

- Will currently fail if _any_ deviation from the standard is detected, which is probably too stringent, need to think of what we consider as "acceptable deviations"
- Do we want to make the Dicom standard configurable per project? [Currently using `"current"`](https://github.com/UCLH-Foundry/PIXL/blob/e6d04cee58274eebc21028616635e517fb42a367/pixl_dcmd/src/pixl_dcmd/main.py#L230), which is also the `dicom_validator` default


pixl_dcmd/tests/test_main.py
```diff
@@ -133,6 +134,10 @@ def test_enforce_whitelist_removes_overlay_plane() -> None:
     assert (0x6000, 0x3000) not in ds
 
 
+def test_validation_check_works(vanilla_dicom_image: Dataset) -> None:
```
stefpiatek, line 137: worth having a test with non-valid dicom

pixl_dcmd/src/pixl_dcmd/main.py
```diff
@@ -216,3 +219,37 @@ def _whitelist_tag(dataset: Dataset, de: DataElement, tag_scheme: list[dict]) ->
...
+    # dicom_validator doesn't have a dedicated API, so setting up the components ourselves. Taken from
+    # https://github.com/pydicom/dicom-validator/blob/37a5f9770ae8f44fa6fb2aaa6bfed57743e01896/dicom_validator/validate_iods.py
+    edition_reader = EditionReader(standard_path)
+    destination = edition_reader.get_revision(revision, False)
+    json_path = Path(destination, "json")
+    dicom_info = EditionReader.load_dicom_info(json_path)
+
+    # NOTE: dicom_validator returns a dict with all deviations from the standard
+    # Currently we're only letting datasets pass that don't have any exceptions
+    # This is likely to stringent for most applications, so we might want to parse the results more
```
stefpiatek, line 241: Would be interesting to see if this is the case on some live data. I'd expect we want to be fully compliant with DICOM spec. Could use this as a way to onboard onto running on the GAE as well
```suggestion
    # This is likely too stringent for most applications, so we might want to parse the results more
```

stefpiatek (approved):
Looks good, let's try this on the GAE before we decide whether we want to alter compliance etc.

milanmlft:
> Looks good, let's try this on the GAE before we decide whether we want to alter compliance etc.

Yeah, but I suspect there will be quite some errors from `dicom-validator`, as it doesn't distinguish (at least not in an easily parsable way) between really critical deviations and mild "warnings".

For example, running it on our test Dicom files (`test/resources/Dicom1.dcm`), returns

```
❯ validate_iods Dicom1.dcm
Using DICOM revision 2024b

Processing DICOM file "Dicom1.dcm"
SOP class is "1.2.840.10008.5.1.4.1.1.1.1" (Digital X-Ray Image IOD)

Errors
======
General:
Tag (0018,9332) (Exposure in mAs) is unexpected
```

Not sure what the reason is here, but I'd assume that an "unexpected tag" is not as critical as a missing or misformatted tag? @HChughtai might have more thoughts.

HChughtai (approved):
Looks good to me as a first pass that rejects any non-conformities. I agree with your comment that rejecting everything may be too strict, so we may want to relax things in a future change.

HChughtai:
> Not sure what the reason is here, but I'd assume that an "unexpected tag" is not as critical as a missing or misformatted tag? @HChughtai might have more thoughts.

I thought that these "unexpected tag" were warnings that a deprectaed tag was present - e.g. one that was valid in a previous version of the DICOM standard, but has been superceded in the current version. Tags don't get removed in newer stadards to retain backwards compatability, so it's not an error in the same way as a required tag being missing.

That said, I don't know why that one is being flagged. 0018,9332. It's a valid conditionally required tag: https://dicom.innolitics.com/ciods/breast-projection-x-ray-image/enhanced-mammography-image/00189332

HChughtai:
> That said, I don't know why that one is being flagged. 0018,9332. It's a valid conditionally required tag: [dicom.innolitics.com/ciods/breast-projection-x-ray-image/enhanced-mammography-image/00189332](https://dicom.innolitics.com/ciods/breast-projection-x-ray-image/enhanced-mammography-image/00189332)

Sorry that's the wrong IOD. I don't think it's present in the Digital X-Ray Image IOD that @milan's comment has. So it's a DICOM tag that is present, but isn't needed as part of the declared IOD (according to the standard).

milanmlft:
Alright, I updated the DICOM validator so that it only checks the difference in errors between the original dataset and after anonymisation. Unfortunately, it looks like our anonymisation isn't very standard compliant at the moment... Only generating a warning though, so the images will still be sent through, so the system test _should_ pass (in theory). But it's difficult to to properly test the validation after anonymisation.
Looks like we might be entering a world of pain here, or at least some mild discomfort.

pixl_dcmd/src/pixl_dcmd/_dicom_helpers.py
```diff
@@ -35,3 +39,47 @@ def get_project_name_as_string(dataset: Dataset) -> str:
...
+        # Check that the original dataset has been validated
+        try:
+            orig_errors = self.original_errors
+        except AttributeError:
+            raise ValueError("Original dataset not yet validated")
+
+        # Temporarily disable logging to avoid spamming the console
+        logging.disable(logging.ERROR)
+        self.anon_errors = IODValidator(dataset, self.dicom_info).validate()
+        logging.disable(logging.NOTSET)
```
stefpiatek, line 71: interesting, this suggests that we should only allow logging from pixl, rather than logging from all imported modules

pixl_dcmd/tests/test_dicom_validator.py
```diff
@@ -0,0 +1,64 @@
...
+    validator.validate_original(vanilla_dicom_image)
+    anonymise_dicom(vanilla_dicom_image)
+    assert not validator.validate_anonymised(vanilla_dicom_image)
+
+
+def test_validation_fails_after_invalid_tag_modification(vanilla_dicom_image) -> None:
+    """
+    GIVEN a DICOM dataset
+    WHEN an invalid tag operation is performed (e.g. deleting a required tag)
+    THEN validation should fail
```
stefpiatek, line 51: fail probably isn't right now

pixl_dcmd/tests/test_dicom_validator.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+    GIVEN a DICOM dataset
+    WHEN the dataset is validated after anonymisation
+    THEN no errors should be raised
+    """
+    validator = DicomValidator()
+    validator.validate_original(vanilla_dicom_image)
+    anonymise_dicom(vanilla_dicom_image)
+    assert not validator.validate_anonymised(vanilla_dicom_image)
+
```
stefpiatek, line 45: it'd be good to have an example of a DICOM which starts off non-compliant, and is anonymised to no warnings

stefpiatek (approved):
Looks nice, as you said, need to work out why CI is failing but happy for you to merge once you've got that and happy with the other suggestions

milanmlft:
> Looks nice, as you said, need to work out why CI is failing but happy for you to merge once you've got that and happy with the other suggestions

Right, but the `pixl_dcmd` tests are not going to pass at the moment, unless we make our anonymisation standard-compliant, or I remove the tests :/ so what do we want to do?

PR #416, opened 2024-07-02
Process large studies and prod fixes

### LINKED ISSUE #395: Avoid unknown projects being sent to orthanc-anon

as a pixl pipeline runner, I don't want to see logs of unknown projects in orthanc-anon, or waste resources sending messages we don't know how to anonymise. 

- [x] When a study is stable, we shouldn't export DICOM data that has the default value of `PrivateDicomTag.unknown_value` 

### Testing

_No response_

### Documentation

_No response_

### Dependencies

_No response_

### Details and Comments

_No response_

stefpiatek:
Had a whole lot of fun with running in production:

- Tweaking of logging
- I _think_ if the GAE has a lot of CPU use then the resource modification can freeze up, this decreases if we don't limit the number of threads that can be used for this so
- Tried cancelling jobs on failure, but it doesn't actually cancel them so removed
- Actually use tqdm properly :facepalm:
- Use DICOM transfer timeout across services in case of large studies
- Increase concurrent jobs in orthanc anon as we have done in orthanc raw
- Make the stable age in orthanc raw 60 seconds longer than the timeout, we were deleting studies that hung after 60 seconds and then we'd get more data and export the latter half of a study!
- Use http timeout for export api, get the status back of the export in case it fails - could make an issue to convert this to async job and check that job status?
- Delete a study if it is stable but has not had the default pixl project tag overriden in orthanc raw, as a way to gate exit of closes #395 
- Be less busy in polling for job status in imaging api, every 10 seconds is fine

TODO:

- [x] Document changes from the end users perspective in README
- [x] Document suggested values for configuration and alter the defaults to match production
- [x] Increase test coverage?


docker-compose.yml
```diff
@@ -149,7 +150,6 @@ services:
             - type: volume
               source: orthanc-anon-data
               target: /var/lib/orthanc/db
-            - ${PWD}/orthanc/orthanc-anon/config:/run/secrets:ro
```
milanmlft, line 152: I'm assuming this is because now we're manually copying and modifying the `orthanc.json` in the Dockerfile, but double-checking that this is the case and this was not accidental.

orthanc/orthanc-raw/README.md
milanmlft, line 1: 👌

orthanc/orthanc-raw/plugin/pixl.py
```diff
@@ -58,14 +58,27 @@ def OnChange(changeType, level, resourceId):  # noqa: ARG001
...
     should_auto_route returns true
     """
-    if changeType == orthanc.ChangeType.STABLE_STUDY and should_auto_route():
+    if changeType != orthanc.ChangeType.STABLE_STUDY:
+        return
+    if _get_project_name_from_study_id(resourceId) == DICOM_TAG_PROJECT_NAME.PLACEHOLDER_VALUE:
+        logger.warning("Study {} has not been set with a pixl project name, deleting", resourceId)
+        orthanc.RestApiDelete(f"/studies/{resourceId}")
+        return
+    # delete via rest API and don't autoroute
```
milanmlft, line 67:
```suggestion
    # delete via rest API and don't autoroute
    if _get_project_name_from_study_id(resourceId) == DICOM_TAG_PROJECT_NAME.PLACEHOLDER_VALUE:
        logger.warning("Study {} has not been set with a pixl project name, deleting", resourceId)
        orthanc.RestApiDelete(f"/studies/{resourceId}")
        return
```

Shouldn't the comment go before the relevant code block?

milanmlft (approved):
Looking great! Though could indeed use a bit more test coverage

docker-compose.yml
stefpiatek: yep that's right

orthanc/orthanc-raw/plugin/pixl.py
stefpiatek: it really should, though I should have just deleted the comment because it doesn't tell you why, just reiterates what the code is telling you anyway

PR #419, opened 2024-07-16
Fix tests for `pixl check_env`

### LINKED ISSUE #410: Fix `pixl check-env` tests

The `cli/tests/test_check_env.py`  tests are updated to correctly test the behaviour of `pixl check-env`.

### Testing

Should update `cli/tests/test_check_env.py`

Suggested test cases:

1. `check_env(error=True)` should pass without errors
2. `check_env(error=True)` should return an error when a `NONEXISTENT_VARIABLE` envvar is set in the sample file
3. `check_env(error=False)` should _not_ return an error when a `NONEXISTENT_VARIABLE` envvar is set in the sample file (might be overkill)

### Documentation

_No response_

### Dependencies

_No response_

### Details and Comments

The tests in `test_check_env.py` are currently passing because of the wrong reasons:

1. `test_check_env()` passes because it's calling `check_env` with default options, so `error = False` and so will never return a non-zero exit code
4. `test_check_env_fails()` passes because it's calling `runner.invoke()` in a faulty way. It also has `error=False` so the non-zero exit code doesn't actually come from `check_env`

To reproduce, modifying `test_check_env()` to the following still passes (but shouldn't!):

```python
def test_check_env(tmp_path):
    """
    Test that the check_env command runs without error.
    - check_env works
    - current test env file matches the sample env file
    """
    tmp_sample_env_file = tmp_path / ".env.sample"
    tmp_sample_env_file.write_text("NONEXISTENT_VARIABLE=")

    runner = CliRunner()
    result = runner.invoke(check_env, ["--sample-env-file", tmp_sample_env_file])

    assert result.exit_code == 0
```

p-j-smith:
Fixes #410

- Update tests to invoke `check_env` with `--error`
- Also update `test_check_env_fails` to:
  - pass the sample env file to `check_env`
  - use `pytest.raises` to ensure the test fails for the right reason (i.e. the correction exception is raised)


cli/tests/test_check_env.py
```diff
@@ -28,7 +30,7 @@ def test_check_env():
     - current test env file matches the sample env file
     """
     runner = CliRunner()
-    result = runner.invoke(check_env)
+    result = runner.invoke(check_env, args=["--error"])
```
p-j-smith, line 33: this test is [failing](https://github.com/UCLH-Foundry/PIXL/actions/runs/9956945495/job/27507918188?pr=419#step:5:131) because none of the environment variables in the sample env file are set. How best to set them for the test? Or should we create a test sample env file in the same way `test_check_env_fails` does?

stefpiatek: ah this is #410, I've updated the description to say that this fixes it

stefpiatek (approved):
love to see it

cli/tests/test_check_env.py
p-j-smith: > Can we point to an env file where this has been set, like the system tests one?

oh good idea! I've done that now, although not all variables are set in that file so the rest are now set in the `conftest.py`

> Probably should be one failing test and one passing test rather than two in the same test?

Sorry I don't really follow what you mean?

stefpiatek: this tests feels like a positive and a negative test smooshed together. Feels like what we should do is to have one postitive test
```
result = runner.invoke(check_env, args=["--error"])
```

then another with the negative test 
```
tmp_sample_env_file.write_text("NONEXISTENT_VARIABLE=")

    runner = CliRunner()
    with pytest.raises(UndefinedValueError, match="not found"):
        runner.invoke(
            check_env,
            args=["--error", "--sample-env-file", tmp_sample_env_file],
            catch_exceptions=False,
        )
```

p-j-smith: Ah, I think that's how it currently is (`test_check_env` for the positive, `test_check_env_fails` for the negative):
https://github.com/UCLH-Foundry/PIXL/blob/285d0348444627d4fa7ceaf6232c946fe9ceabfe/cli/tests/test_check_env.py#L26-L50
Is it just the diff collapsing lines and making it look like one test? Sorry if I'm still not following here...

stefpiatek: ha! I could have sworn I checked for that but clearly not. Ignore me

p-j-smith: ha, no worries 😄 thanks for the review!

PR #469, opened 2024-07-31
Use imaging UID from OMOP-ES extract

milanmlft:
## Description
Related to https://github.com/UCLH-Foundry/the-rolling-skeleton/issues/161
I suspect this has to wait until https://github.com/UCLH-Foundry/the-rolling-skeleton/issues/161 is done and the OMOP-ES extracts actually contain the study UIDs.

* Get Study UID from OMOP-ES extract
* Use the Study UID to query the VNA initially, fall back to MRN + accession number if not available
* Add the UID to the PIXL database `Image` model and use for querying during anonymisation

Also cleaned up the `pixl_dcmd` tests and fixtures a bit, as I was getting confused where all the hardcoded values were coming from.

## Type of change

### Suggested Checklist

- [x] I have passed on my local host device. (see further details at the [CONTRIBUTING](https://github.com/UCLH-Foundry/PIXL/blob/main/CONTRIBUTING.md#local-testing) document)
- [x] Make sure your branch is up-to-date with main branch. See [CONTRIBUTING](https://github.com/UCLH-Foundry/PIXL/blob/main/CONTRIBUTING.md) for a general example to syncronise your branch with the `main` branch.
- [x] I have requested PR review to `UCLH-Foundtry/arc-dev`
- [x] I have adddressed and marked as resolved all the review comments in my PR.
- [x] Finally, I have selected `squash and merge`


pixl_dcmd/src/pixl_dcmd/_database.py
```diff
@@ -41,16 +41,15 @@ def get_uniq_pseudo_study_uid_and_update_db(
...
     gets the pseudo_study_uid if it is not None or records a new, unique one.
     Returns the pseudo_study_uid.
     """
     PixlSession = sessionmaker(engine)
     with PixlSession() as pixl_session, pixl_session.begin():
         existing_image = get_unexported_image(
             project_slug,
-            study_info.mrn,
-            study_info.accession_number,
+            study_info.study_uid,
```
stefpiatek, line 52: This method should check by study_uid, and if it can't find it, then search by mrn and accession number to match what we get from the VNA. At that point it throws an exception where it can't find based on either

cli/src/pixl_cli/_io.py
```diff
@@ -160,6 +161,7 @@ class DF_COLUMNS(StrEnum):  # noqa: N801
     "PrimaryMrn": "mrn",
     "AccessionNumber": "accession_number",
     "procedure_date": "study_date",
+    "StudyUid_X": "study_uid",
```
stefpiatek, line 164: might be missing it but we should edit our test CSV files and parquet files so that they have a `study_uid` column

stefpiatek: also seeing it on line `142`, we should also make sure that we filter out any rows where the accession number is an empty string 🙏🏻

pixl_core/src/core/db/models.py
```diff
@@ -52,6 +52,7 @@ class Image(Base):
     accession_number: Mapped[str]
     study_date: Mapped[Date] = mapped_column(Date())
     mrn: Mapped[str]
+    study_uid: Mapped[str]
```
stefpiatek, line 55: You'll also need to make an alembic migration for this change

stefpiatek (approved):
Looking good, UIDs are in omop es, so you can ask Baptiste to set off a pinpoint run again if you want to test on real data

pixl_dcmd/src/pixl_dcmd/_database.py
milanmlft: I started off doing that but then I figured that at this point all images would have a UID known to us, no? Because when we get a DICOM dataset from the VNA, even when queried using MRN + accession number, we can get the UID from the `StudyInstanceUID` field.
Or is it possible for a DICOM instance to lack the `StudyInstanceUID` field?

cli/src/pixl_cli/_io.py
milanmlft: > might be missing it but we should edit our test CSV files and parquet files so that they have a `study_uid` column

Ah yes, I assumed we would get some new test data from an OMOP-ES run. How did we generate the original test data?

pixl_core/src/core/db/models.py
stefpiatek: Oh this also should be optional, or we populate with a dummy value when we create the column

pixl_dcmd/src/pixl_dcmd/_database.py
stefpiatek: Also not only that, but if its an emergency scan then the UID won't match whats in caboodle

milanmlft:
> I wonder for pinpoint we could just delete all rows which haven't been exported and then run populate so the UIDs will exist in the ones its still trying to export. From thereon in, we wouldn't expect the UID to change so that should be fine

Sounds reasonable!

PR #476, opened 2024-08-07
Set development version

### LINKED ISSUE #473: Bump module versions to `v0.1.0` and create release
This will be the version of PIXL that will be used for the PINPOINT project

- [x] Bump version for each module to `0.1.0
- [x] Add `0.1.0` tag
- [x] Create GitHub release and write some release notes
- [ ] Bump modules to `0.2.0-rc` after release

milanmlft:
* Bump module versions to `0.2.0-rc` following the [v0.1.0 release](https://github.com/UCLH-Foundry/PIXL/releases/tag/v0.1.0)
* Update dependency pins for `pixl_core` and `pytest_pixl` to `>=0.1.0`

Closes #473

### DIFF
--- a/cli/pyproject.toml
+++ b/cli/pyproject.toml
@@ -7,7 +7,7 @@ readme = "README.md"
 requires-python = ">=3.10"
 classifiers = ["Programming Language :: Python :: 3"]
 dependencies = [
-    "core==0.1.0",
+    "core>=0.1.0",
     "click==8.1.7",
     "tqdm==4.66.4",
 ]
--- a/cli/pyproject.toml
+++ b/cli/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "pixl_cli"
-version = "0.1.0"
+version = "0.2.0-rc"
 authors = [{ name = "PIXL authors" }]
 description = "PIXL command line interface"
 readme = "README.md"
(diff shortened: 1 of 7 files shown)


cli/pyproject.toml
```diff
@@ -1,13 +1,13 @@
...
-version = "0.1.0"
+version = "0.2.0-rc"
 authors = [{ name = "PIXL authors" }]
 description = "PIXL command line interface"
 readme = "README.md"
 requires-python = ">=3.10"
 classifiers = ["Programming Language :: Python :: 3"]
 dependencies = [
-    "core==0.1.0",
+    "core>=0.1.0",
```
p-j-smith, line 10: should this be pinned rather than a minimum version?

p-j-smith (approved):
very nice, just a small suggestion around the version naming

cli/pyproject.toml
stefpiatek: Yeah I'd probably pin it to the rc

stefpiatek (approved):
Thanks for setting this up! Not much to add on top of Paul's review

milanmlft:
Thanks both! Regarding the dependency pinning, I guess this comes from my R dev background, where it's more common to set minimum versions - which in my mind also makes more sense... but I will concede to the Pythonic way 😁

PR #489, opened 2024-08-22
Add support for querying PACS

### LINKED ISSUE #465: Fix buggy system test in CI

- [ ] Understand the bug and fix it (see details)
- [ ] Try running action a few times in CI to make sure that doesn't crop up

### Testing

_No response_

### Documentation

_No response_

### Dependencies

_No response_

### Details and Comments

With the exact same version of the code the system tests sometimes have a failure. Haven't looked but I think it might be fixed by checking `('ANONYMIZED', 'include123')` is in the set, rather than the two sets are equal. Would need to check that makes sense for the requirements of the test

```
 FAILED system_test.py::TestFtpsUpload::test_ftps_dicom_upload - AssertionError: assert {('ANONYMIZED', 'include123')} == {('ANONYMIZED...'include123')}
  
  Extra items in the right set:
  ('ANONYMIZED', 'AP')
  
  Full diff:
    {
  -     (
  -         'ANONYMIZED',
  -         'AP',
  -     ),
        (
            'ANONYMIZED',
            'include123',
        ),
    }
```

[passing action](https://github.com/UCLH-Foundry/PIXL/actions/runs/10178090427/job/28151809671?pr=464)
[failing action](https://github.com/UCLH-Foundry/PIXL/actions/runs/10178090427/job/28151036968?pr=464)

### LINKED ISSUE #480: Allow for secondary dicom server querying

As a researcher I want to be able to query PACS if we can't find the data in the VNA

- [x] First try VNA, fallback to PACS if this isn't in the VNA
  - [x] to start with try only 8pm-8am monday-friday to ensure the dicom server 
  - [x] Query for [instance availablity](https://dicom.nema.org/medical/dicom/2020b/output/chtml/part03/sect_C.4.23.html ) in the c-find, only pull from PACS if it is online (nearline will probably be routing from the VNA), otherwise log that that it should be available in the primary DICOM server

### Testing

- [x] Automated test that has an image only available in secondary and it should be found and processed

### Documentation

- [x] Update env template
- [x] Update documenation of dicom processing 

### Dependencies

_No response_

### Details and Comments

Will probably want to rename the VNA variables to be PRIMARY_DICOM, and then PACS can be the secondary dicom

p-j-smith:
## Description
Fixes #480 

- query the VNA for a study, trying PACS instead if a study is not found
- if the study is found in PACS but it's available online, do not pull the study (we're assuming it's available in the VNA)
- add test to check a study is pulled from PACS if it's not available in the VNA
- update imaging API docs to explain the secondary dicom archive is queried if a study is not in the primary dicom archive
- rename `VNAQR_` environment variables to `PRIMARY_DICOM_`
- add environment variables for `SECONARY_DICOM_` (PACS)

Fixes #465 
- increase `ORTHANC_RAW_EXTRA_STABLE_SECONDS` and `PIXL_DICOM_TRANSFER_TIMEOUT` 30 seconds for the system tests to ensure the study is stable (i.e. all series are present)

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

### Suggested Checklist

- [x] I have passed on my local host device. (see further details at the [CONTRIBUTING](https://github.com/UCLH-Foundry/PIXL/blob/main/CONTRIBUTING.md#local-testing) document)
- [x] Make sure your branch is up-to-date with main branch. See [CONTRIBUTING](https://github.com/UCLH-Foundry/PIXL/blob/main/CONTRIBUTING.md) for a general example to syncronise your branch with the `main` branch.
- [x] I have requested PR review to `UCLH-Foundtry/arc-dev`
- [ ] I have adddressed and marked as resolved all the review comments in my PR.


pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -143,27 +144,112 @@ async def _add_project_to_study(
...
+    study: ImagingStudy,
+    modality: str,
+) -> Optional[str]:
+    """
+    Query the primary archive for the study using its UID.
+    If UID is not available, query on MRN and accession number.
+    """
+    # We don't want to normalize the query otherwise only MainDicomTags will be returned
+    # (InstanceAvailability will be ignored)
+    additional_data = {"Normalize": False}
```
p-j-smith, line 219: I've not add `"Normalize"` to `ImagingStudy.orthanc_uid_query_dict` as this caused an error with

pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -143,27 +144,112 @@ async def _add_project_to_study(
...
+        raise PixlDiscardError(msg)
+
+    # Check the study is in the online secondary archive
+    query_answers = await orthanc_raw.get_remote_query_answers(query_id)
+    query_answer_content = await orthanc_raw.get_remote_query_answer_content(
+        query_id=query_id,
+        answer_id=query_answers[0],
+    )
+    availablility_tag = "0008,0056"
+    availability = query_answer_content[availablility_tag]["Value"]
```
p-j-smith, line 197: will the instance availability always be present?

stefpiatek:
oh but it came up again, okay worth fixing as well

orthanc/orthanc-raw/README.md
```diff
@@ -43,7 +43,7 @@ secrets. Orthanc interprets all `.json` files in the `/run/secrets` mount as con
 
 ### Step 1
 
-Save credentials `.env` for the PACS/VNA Q/R target, postgreSQL and 'Orthanc anon'.
+Save credentials `.env` for the VNA (primary) and PACS (secondary) Q/R targets, postgreSQL and 'Orthanc anon'.
```
stefpiatek, line 46: might be worth saying that the secondary is optional, and if it is not set then only the primary will be used?

orthanc/orthanc-raw/config/dicom.json
```diff
@@ -3,9 +3,25 @@
 
   "DicomModalities" : {
     "UCVNAQR" : {
```
stefpiatek, line 5: I wonder if we should change these key names to be primary and secondary?

pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -143,27 +144,112 @@ async def _add_project_to_study(
...
-            study.orthanc_query_dict, modality=config("VNAQR_MODALITY")
+
+    if query_id is not None:
+        return str(query_id)
+    if _is_daytime() or _is_weekend():
+        msg = (
+            f"Failed to find study {study.message.study_uid} in primary archive. "
+            "Not querying secondary archive during the daytime or on the weekend."
         )
+        raise PixlDiscardError(msg)
```
stefpiatek, line 174: ah we should have a test that covers this, presumably using freezegun or something like that to fix the date and time

pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -143,27 +144,112 @@ async def _add_project_to_study(
...
+    query_id = await _find_study_in_archive(
+        study=study,
+        orthanc_raw=orthanc_raw,
+        modality=config("SECONDARY_DICOM_MODALITY"),
+    )
+
     if query_id is None:
-        msg = "Failed to find in the VNA"
+        msg = f"Failed to find study {study.message.study_uid} in primary or secondary archive."
+        raise PixlDiscardError(msg)
```
stefpiatek, line 188: Oh do we have a test case for this, or is coverage playing up again?

pixl_imaging/src/pixl_imaging/_processing.py
stefpiatek: I think so, and if not then we want it to fail so assuming that it is seems reasonable

pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -143,27 +144,112 @@ async def _add_project_to_study(
...
+) -> Optional[str]:
+    """
+    Query the primary archive for the study using its UID.
+    If UID is not available, query on MRN and accession number.
+    """
+    # We don't want to normalize the query otherwise only MainDicomTags will be returned
+    # (InstanceAvailability will be ignored)
+    additional_data = {"Normalize": False}
+    query_response = await orthanc_raw.query_remote(
+        data=study.orthanc_uid_query_dict | additional_data,
```
stefpiatek, line 221: oh this is fun

pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -143,27 +144,112 @@ async def _add_project_to_study(
...
+        study.orthanc_query_dict | additional_data,
+        modality=modality,
+    )
+
+
+def _is_daytime() -> bool:
+    """Check if the current time is between 8 am and 8 pm."""
+    after_8am = datetime.time(8, 00) <= datetime.datetime.now(tz=datetime.UTC).time()
+    before_8pm = datetime.datetime.now(tz=datetime.UTC).time() <= datetime.time(20, 00)
+    return after_8am and before_8pm
```
stefpiatek, line 243: I wonder if we should be doing this in local time and setting the timezone explicitly to the `London/Europe` in all our docker containers

pixl_imaging/tests/test_imaging_processing.py
```diff
@@ -112,6 +167,7 @@ async def orthanc_raw(run_containers) -> PIXLRawOrthanc:
 @pytest.mark.processing()
 @pytest.mark.asyncio()
 @pytest.mark.usefixtures("_add_image_to_fake_vna")
+@pytest.mark.usefixtures("_add_image_to_fake_pacs")
```
stefpiatek, line 170: do we need to use this fixture here? and on L189 and L217?

pixl_imaging/tests/test_imaging_processing.py
```diff
@@ -171,6 +229,107 @@ async def test_querying_without_uid(orthanc_raw, caplog) -> None:
...
+    """Mock the date and time to be a Monday at 2 am."""
+
+    @classmethod
+    def now(cls, tz=None) -> datetime.datetime:
+        return MockDateTime(2024, 1, 1, 2, 0, tzinfo=tz)
+
+
+@pytest.mark.processing()
+@pytest.mark.asyncio()
+@pytest.mark.usefixtures("_add_image_to_fake_vna")
```
stefpiatek, line 248: can probably remove using this fixture so that the VNA has no data in it at all. I'm assuming that the teardown here removes the data from the VNA?

pixl_imaging/tests/test_imaging_processing.py
```diff
@@ -171,6 +229,107 @@ async def test_querying_without_uid(orthanc_raw, caplog) -> None:
...
+        "trying MRN and accession number"
+    )
+    assert unexpected_msg not in caplog.text
+
+
+@pytest.mark.processing()
+@pytest.mark.asyncio()
+@pytest.mark.usefixtures("_add_image_to_fake_vna")
+@pytest.mark.usefixtures("_add_image_to_fake_pacs")
+async def test_querying_pacs_without_uid(orthanc_raw, caplog, monkeypatch) -> None:
```
stefpiatek, line 296: also would be good to have a test to show that during working hours, the PACS is not queried

pixl_imaging/tests/test_imaging_processing.py
```diff
@@ -171,6 +229,107 @@ async def test_querying_without_uid(orthanc_raw, caplog) -> None:
...
+    """
+    Given a message with non-existent study_uid exists in PACS but not VNA,
+    When we query the archives
```
stefpiatek, line 299: how difficult would it be to set up a test to see that if the secondary dicom isn't defined, then it doesn't try querying it?

pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -143,27 +144,112 @@ async def _add_project_to_study(
...
+
+    if query_id is not None:
+        return str(query_id)
+    if _is_daytime() or _is_weekend():
+        msg = (
+            f"Failed to find study {study.message.study_uid} in primary archive. "
+            "Not querying secondary archive during the daytime or on the weekend."
         )
+        raise PixlDiscardError(msg)
+
```
stefpiatek, line 175: should also return early if the secondary dicom server isn't defined (say AE title?), so that we don't have to use the PACS

stefpiatek (approved):
Looking good! I think the only thing that on thinking about it would be useful (I didn't say on the issue) is that if the secondary dicom server hasn't been defined then we should return early

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: ooo yeah freezegun looks nice, I might switch to that

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: yeah definitely, that's a good shout

pixl_imaging/tests/test_imaging_processing.py
p-j-smith: oh probably not, well spotted

pixl_imaging/tests/test_imaging_processing.py
p-j-smith: ah yeah that would definitely make testing easier

p-j-smith:
> I think the only thing that on thinking about it would be useful (I didn't say on the issue) is that if the secondary dicom server hasn't been defined then we should return early

Oh good idea, what's the best way to check this? Check that that `SECONDARY_DICOM_MODALITY` environment variable is defined and that the modality exists on Orthanc?

stefpiatek:
> Oh good idea, what's the best way to check this? Check that that `SECONDARY_DICOM_MODALITY` environment variable is defined and that the modality exists on Orthanc?

Yeah I think so. Though I'm not sure what we do on the orthanc configuration side, maybe set some default dummy values as it'll never be connected to?

test/.env
mxochicale, line 1: I have run `pytest -vv test` from root path of PIXL and I got this. Am I missing anything to pass the tests? See few lines of the initial log error.

```
test/system_test.py::TestFtpsUpload::test_ftps_parquet_upload ERROR                                                                                                                                         [ 20%]
test/system_test.py::TestFtpsUpload::test_ftps_radiology_linker_upload ERROR                                                                                                                                [ 40%]
test/system_test.py::TestFtpsUpload::test_ftps_dicom_upload ERROR                                                                                                                                           [ 60%]
test/system_test.py::test_dicomweb_upload ERROR                                                                                                                                                             [ 80%]
test/test_parquet_exports.py::test_public_parquet ERROR                                                                                                                                                     [100%]

===================================================================================================== ERRORS ======================================================================================================
____________________________________________________________________________ ERROR at setup of TestFtpsUpload.test_ftps_parquet_upload ____________________________________________________________________________

self = <urllib3.connection.HTTPConnection object at 0x74831deeb910>

    def _new_conn(self) -> socket.socket:
        """Establish a socket connection and set nodelay settings on it.
    
        :return: New socket connection.
        """
        try:
>           sock = connection.create_connection(
                (self._dns_host, self.port),
                self.timeout,
                source_address=self.source_address,
                socket_options=self.socket_options,
            )

../../../../mambaforge/envs/pixlVE/lib/python3.11/site-packages/urllib3/connection.py:196: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../../mambaforge/envs/pixlVE/lib/python3.11/site-packages/urllib3/util/connection.py:85: in create_connection
    raise err
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

address = ('localhost', 8043), timeout = 60, source_address = None, socket_options = [(6, 1, 1)]

    def create_connection(
        address: tuple[str, int],
        timeout: _TYPE_TIMEOUT = _DEFAULT_TIMEOUT,
        source_address: tuple[str, int] | None = None,
        socket_options: _TYPE_SOCKET_OPTIONS | None = None,
    ) -> socket.socket:
        """Connect to *address* and return the socket object.
    
        Convenience function.  Connect to *address* (a 2-tuple ``(host,
        port)``) and return the socket object.  Passing the optional
        *timeout* parameter will set the timeout on the socket instance
        before attempting to connect.  If no *timeout* is supplied, the
        global default timeout setting returned by :func:`socket.getdefaulttimeout`
        is used.  If *source_address* is set it must be a tuple of (host, port)
        for the socket to bind as a source address before making the connection.
        An host of '' or port 0 tells the OS to use the default.
        """
    
        host, port = address
        if host.startswith("["):
            host = host.strip("[]")
        err = None
    
        # Using the value from allowed_gai_family() in the context of getaddrinfo lets
        # us select whether to work with IPv4 DNS records, IPv6 records, or both.
        # The original create_connection function always returns all records.
        family = allowed_gai_family()
    
        try:
            host.encode("idna")
        except UnicodeError:
            raise LocationParseError(f"'{host}', label empty or too long") from None
    
        for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
            af, socktype, proto, canonname, sa = res
            sock = None
            try:
                sock = socket.socket(af, socktype, proto)
    
                # If provided, set socket level options before connecting.
                _set_socket_options(sock, socket_options)
    
                if timeout is not _DEFAULT_TIMEOUT:
                    sock.settimeout(timeout)
                if source_address:
                    sock.bind(source_address)
>               sock.connect(sa)
E               ConnectionRefusedError: [Errno 111] Connection refused

../../../../mambaforge/envs/pixlVE/lib/python3.11/site-packages/urllib3/util/connection.py:73: ConnectionRefusedError

The above exception was the direct cause of the following exception:

self = <urllib3.connectionpool.HTTPConnectionPool object at 0x74831c0f4090>, method = 'POST', url = '/instances'
body = b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00...00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
headers = {'User-Agent': 'python-requests/2.32.2', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive', 'Content-Length': '527726', 'Authorization': 'Basic b3J0aGFuYzpvcnRoYW5j'}
retries = Retry(total=0, connect=None, read=False, redirect=None, status=None), redirect = False, assert_same_host = False, timeout = Timeout(connect=60, read=60, total=None), pool_timeout = None
release_conn = False, chunked = False, body_pos = None, preload_content = False, decode_content = False, response_kw = {}
parsed_url = Url(scheme=None, auth=None, host=None, port=None, path='/instances', query=None, fragment=None), destination_scheme = None, conn = None, release_this_conn = True, http_tunnel_required = False
err = None, clean_exit = False

    def urlopen(  # type: ignore[override]
        self,

```

mxochicale (approved):
It looks good to me. Not too much to add for me at this stage.  I have just tested `pytest -vv pixl_imaging/tests/` and 5 tests passed but I bumped into an error that might be related to some missing steps from my side.

test/.env
stefpiatek: I think this was the error I encountered, needing to turn default sockets in docker compose config?
on a mac I can find this setting in  Settings > Advanced
image

mxochicale: I don't see that setting in `docker deskop` for the linux version (see below screenshot). Is there any way to setup manually?  do I need to sing in? 

Screenshot from 2024-08-27 13-34-35

stefpiatek: googling around I found this, which might require editing of the docker.service file? https://scriptcrunch.com/enable-docker-remote-api/

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: Ah good point, I've added a test for it

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: yep!

orthanc/orthanc-raw/config/dicom.json
p-j-smith: I've renamed them to `UCPRIMARYQR` and `UCSECONDARYQR`

pixl_imaging/tests/test_imaging_processing.py
p-j-smith: I've removed the unused fixtures

pixl_imaging/tests/test_imaging_processing.py
p-j-smith: I've removed these unused fixtures too

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: The timezone if already set in the [`common.env`](https://github.com/UCLH-Foundry/PIXL/blob/main/docker/common.env). I've also set `TZ` for the containers than don't use this common env file. I've also installed `tzdata` just incase it's missing from the images

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: I've switched to freezegun now (much nicer)

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: https://github.com/UCLH-Foundry/PIXL/blob/0de17bb6da2f16b43e6f7ae50b745e3488d24130/pixl_imaging/tests/test_imaging_processing.py#L333-L350

pixl_imaging/tests/test_imaging_processing.py
p-j-smith: yeah definitely, I've added tests [here](https://github.com/UCLH-Foundry/PIXL/blob/0de17bb6da2f16b43e6f7ae50b745e3488d24130/pixl_imaging/tests/test_imaging_processing.py#L353-L381)

pixl_imaging/tests/test_imaging_processing.py
p-j-smith: Yeah a test would be good, but I'm not sure how best to test it. I guess we could have a function that returns the list of modalities in orthanc, and mock it in the test to not return pacs?

stefpiatek: Could we make it look for the environmental variable within the function that runs, and then if its not defined at that time, then it doesn't use it? That way we could set the environmental variable for the setup of just that test I think

pixl_imaging/src/pixl_imaging/_processing.py
stefpiatek: ah even better

test/.env
mxochicale: From the first link, just edited `vim usr/lib/systemd/system/docker.service`: 
```
#ExecStart=/usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock
ExecStart=/usr/bin/dockerd -H tcp://0.0.0.0:4243 -H unix:///var/run/docker.sock
```
and `sudo systemctl daemon-reload` `sudo service docker restart`  but still got the following logs. 

```
                                                                                                                                         

test/system_test.py::TestFtpsUpload::test_ftps_parquet_upload ERROR                                                                                                                                         [ 20%]
test/system_test.py::TestFtpsUpload::test_ftps_radiology_linker_upload ERROR                                                                                                                                [ 40%]
test/system_test.py::TestFtpsUpload::test_ftps_dicom_upload ERROR                                                                                                                                           [ 60%]
test/system_test.py::test_dicomweb_upload ERROR                                                                                                                                                             [ 80%]
test/test_parquet_exports.py::test_public_parquet ERROR                                                                                                                                                     [100%]

===================================================================================================== ERRORS ======================================================================================================
____________________________________________________________________________ ERROR at setup of TestFtpsUpload.test_ftps_parquet_upload ____________________________________________________________________________

self = <urllib3.connection.HTTPConnection object at 0x71ea48d1e190>

    def _new_conn(self) -> socket.socket:
        """Establish a socket connection and set nodelay settings on it.
    
        :return: New socket connection.
        """
        try:
>           sock = connection.create_connection(
                (self._dns_host, self.port),
                self.timeout,
                source_address=self.source_address,
                socket_options=self.socket_options,
            )

../../../../mambaforge/envs/pixlVE/lib/python3.11/site-packages/urllib3/connection.py:196: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../../../../mambaforge/envs/pixlVE/lib/python3.11/site-packages/urllib3/util/connection.py:85: in create_connection
    raise err
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

address = ('localhost', 8043), timeout = 60, source_address = None, socket_options = [(6, 1, 1)]

    def create_connection(
        address: tuple[str, int],
        timeout: _TYPE_TIMEOUT = _DEFAULT_TIMEOUT,
        source_address: tuple[str, int] | None = None,
        socket_options: _TYPE_SOCKET_OPTIONS | None = None,
    ) -> socket.socket:
        """Connect to *address* and return the socket object.
    
        Convenience function.  Connect to *address* (a 2-tuple ``(host,
        port)``) and return the socket object.  Passing the optional
        *timeout* parameter will set the timeout on the socket instance
        before attempting to connect.  If no *timeout* is supplied, the
        global default timeout setting returned by :func:`socket.getdefaulttimeout`
        is used.  If *source_address* is set it must be a tuple of (host, port)
        for the socket to bind as a source address before making the connection.
        An host of '' or port 0 tells the OS to use the default.
        """
    
        host, port = address
        if host.startswith("["):
            host = host.strip("[]")
        err = None
    
        # Using the value from allowed_gai_family() in the context of getaddrinfo lets
        # us select whether to work with IPv4 DNS records, IPv6 records, or both.
        # The original create_connection function always returns all records.
        family = allowed_gai_family()
    
        try:
            host.encode("idna")
        except UnicodeError:
            raise LocationParseError(f"'{host}', label empty or too long") from None
    
        for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
            af, socktype, proto, canonname, sa = res
            sock = None
            try:
                sock = socket.socket(af, socktype, proto)
    
                # If provided, set socket level options before connecting.
                _set_socket_options(sock, socket_options)
    
                if timeout is not _DEFAULT_TIMEOUT:
                    sock.settimeout(timeout)
                if source_address:
                    sock.bind(source_address)
>               sock.connect(sa)
E               ConnectionRefusedError: [Errno 111] Connection refused

../../../../mambaforge/envs/pixlVE/lib/python3.11/site-packages/urllib3/util/connection.py:73: ConnectionRefusedError

The above exception was the direct cause of the following exception:

self = <urllib3.connectionpool.HTTPConnectionPool object at 0x71ea5cd8a890>, method = 'POST', url = '/instances'
body = b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00...00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
headers = {'User-Agent': 'python-requests/2.32.2', 'Accept-Encoding': 'gzip, deflate', 'Accept': '*/*', 'Connection': 'keep-alive', 'Content-Length': '527726', 'Authorization': 'Basic b3J0aGFuYzpvcnRoYW5j'}
retries = Retry(total=0, connect=None, read=False, redirect=None, status=None), redirect = False, assert_same_host = False, timeout = Timeout(connect=60, read=60, total=None), pool_timeout = None
release_conn = False, chunked = False, body_pos = None, preload_content = False, decode_content = False, response_kw = {}
parsed_url = Url(scheme=None, auth=None, host=None, port=None, path='/instances', query=None, fragment=None), destination_scheme = None, conn = None, release_this_conn = True, http_tunnel_required = False
err = None, clean_exit = False


```

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: ah I see, I've done this now, rasing an error if `SECONDARY_DICOM_AE_TITLE` is undefined / an empty string

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: I've switched back from freezegun as it also froze `time.time`, which meant the processing would always hang [here](https://github.com/SAFEHR-data/PIXL/blob/9e082f112000f684cf6d854d62cc35904e444f15/pixl_imaging/src/pixl_imaging/_orthanc.py#L134). I tried setting `ignore=["time"]` or `ignore=["time.time"]` in the call to `freeze_time` but my ignore list was ignored...

test/.env
p-j-smith: @mxochicale shall we take a look together when we pair this afternoon?

mxochicale: Thanks Paul and Stef

Indeed, to make `docker ps` work again, I just need to reboot my machine after making this changes `ExecStart=/usr/bin/dockerd -H fd:// --containerd=/run/containerd/containerd.sock #ExecStart=/usr/bin/dockerd -H tcp://0.0.0.0:4243 -H unix:///var/run/docker.sock` in `usr/lib/systemd/system/docker.service.` 

I have just run `run-system-test.sh` at tests path, but I got the following logs summary. Let's catch tomorrow.
```
================================================================================== short test summary info ===================================================================================
ERROR system_test.py::TestFtpsUpload::test_ftps_parquet_upload - AssertionError: Condition at_least_n_studies_exported was not achieved even after 181 seconds
ERROR system_test.py::TestFtpsUpload::test_ftps_radiology_linker_upload - AssertionError: Condition at_least_n_studies_exported was not achieved even after 181 seconds
ERROR system_test.py::TestFtpsUpload::test_ftps_dicom_upload - AssertionError: Condition at_least_n_studies_exported was not achieved even after 181 seconds
ERROR test_parquet_exports.py::test_public_parquet - AssertionError: Condition at_least_n_studies_exported was not achieved even after 181 seconds
========================================================================== 1 passed, 4 errors in 310.87s (0:05:10) ===========================================================================

```

orthanc/orthanc-raw/config/dicom.json
```diff
@@ -2,10 +2,26 @@
...
+      "AllowStore" : true,
+      "AllowStorageCommitment" : false,
+      "AllowTranscoding" : true,
+      "UseDicomTls" : false,
+      "Timeout" : 60
+    },
+    "UCSECONDARYQR" : {
+      "AET" : "${SECONDARY_DICOM_AE_TITLE}",
+      "Port" : "${SECONDARY_DICOM_PORT}",
+      "Host" : "${SECONDARY_DICOM_IP_ADDR}",
```
stefpiatek, line 24: what happens if we don't define the secondary QR in a deployment, I'm hoping that we just can't use that QR but it orthanc raw still runs?

If not, we could have the default values to be the same as the primary, and check to see if the primary and secondary QR are the same, and then do the early exit there?

pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -143,27 +145,122 @@ async def _add_project_to_study(
...
-    UID is not available, raise exception if it doesn't exist
+    Query primary and secondary archives for a study.
+
+    - Query the primary archive for the study using its UID. If UID is not available, query on
+      MRN and accession number.
+    - If not found in the primary archive, and it is a weekday between 8 pm and 8 am, query the
+      online secondary archive using the study UID. If UID is not available, query on
+      MRN + accession number.
+    - If not found in the online secondary archive, query the nearline secondary archive. If found
+      in the nearline, assume the study is in the primary archive and log it.
```
stefpiatek, line 158: I'd probably say that we should have a discard error again here, with the cause being that it its in the nearline

pixl_imaging/src/pixl_imaging/_processing.py
stefpiatek: ha sods law, yeah the time mocking isn't super painful so seems reasonable

pixl_imaging/src/pixl_imaging/_processing.py
stefpiatek: yeah reasonable way around this

pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -143,27 +145,122 @@ async def _add_project_to_study(
...
+    """Check if the current time is between 8 am and 8 pm."""
+    timezone = ZoneInfo(config("TZ", default="Europe/London"))
+    after_8am = datetime.time(8, 00) <= datetime.datetime.now(tz=timezone).time()
+    before_8pm = datetime.datetime.now(tz=timezone).time() <= datetime.time(20, 00)
+    return after_8am and before_8pm
+
+
+def _is_weekend() -> bool:
+    """Check if it's the weekend."""
+    timezone = ZoneInfo(config("TZ", default="Europe/London"))
```
stefpiatek, line 258: I guess we should think about where we want to define the defaults. Currently we've got them in the docker envs and in the code. Don't particularly mind where but the default should only be in one place IMHO

stefpiatek (approved):
Nice! Hopefully this can be deployed as-is without a secondary dicom server being defined. I also wonder if its worth calling them PRIMARY_DICOM_SOURCE_* in the envs to make it clear that we're pulling data from them?

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: oh yeah that makes more sense

orthanc/orthanc-raw/config/dicom.json
p-j-smith: oh good point - I just tried it with the system tests and orthanc raw failed to start. Yeah defaulting to same as primary it a good idea

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: yeah that would be much neater. I guess it makes sense to set it in docker

orthanc/orthanc-raw/config/dicom.json
p-j-smith: the `SECONDARY_DICOM_SOURCE_` vars now [default](https://github.com/SAFEHR-data/PIXL/blob/6014fcd4f15957cf3e86c62acb49b6b1279ab406/docker-compose.yml#L200-L202) to `PRIMARY_DECOM_SOURCE_` values:

pixl_imaging/src/pixl_imaging/_processing.py
p-j-smith: I've removed the defaults here and only set them in the docker compose configs

p-j-smith:
Thanks for the reviews! I think the systems tests should be passing locally now @mxochicale but let me know if you're still having issues

PR #495, opened 2024-09-03
Check the study UID in Orthanc Raw is the same as in VNA / PACS [force-system-test]

p-j-smith:
## Description
Related to #486.
- For #486, we need to compare the `SOPInstanceUID`s in Orthanc Raw against those in the VNA / PACS. However, when retrieving data, the `StudyInstanceUID`, and `SeriesInstanceUID`, and `SOPInstanceUID` values are modified by Orthanc Raw.

Fixes #496 
- keep `StudyInstanceUID`, `SeriesInstanceUID`, and `SOPInstanceUID` when setting the project name private tag in Orthanc Raw. Also set `Force=True` to allow us to keep these UIDs
- set `OverwriteInstance: True` in the PIXL Raw config so that if an instance with an existing SOPInstanceUID is uploaded, the previous instance is overwritten (otherwise we'd need to manually delete the old instance after setting the project name tag)
- Add test to check the `StudyInstanceUID`, `SeriesInstanceUID`, and `SOPInstanceUID` are correct in Orthanc raw after retrieving a study and setting the project name tag
- update the config for Orthanc Raw in the imaging API tests to be the same as the production config

## Type of change
Please delete options accordingly to the description.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

### Suggested Checklist

- [x] I have performed a self-review of my own code.
- [ ] I have made corresponding changes to the documentation.
- [ ] My changes generate no new warnings.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have passed on my local host device. (see further details at the [CONTRIBUTING](https://github.com/SAFEHR-data/PIXL/blob/main/CONTRIBUTING.md#local-testing) document)
- [ ] Make sure your branch is up-to-date with main branch. See [CONTRIBUTING](https://github.com/SAFEHR-data/PIXL/blob/main/CONTRIBUTING.md) for a general example to syncronise your branch with the `main` branch.
- [ ] I have requested PR review to `UCLH-Foundtry/arc-dev`
- [ ] I have adddressed and marked as resolved all the review comments in my PR.
- [ ] Finally, I have selected `squash and merge`

PR #497, opened 2024-09-03
Recover from partial DICOM transfers from archives to Orthanc Raw

### LINKED ISSUE #486: Recover from partial transfers of DICOM from source

Currently the pipeline deletes a study if it becomes stable but doesn't have the custom DICOM tag for pixl project set. We should allow for a graceful recovery from a DICOM that failed transfer midway, only downloading the data that hasn't previously been transferred.

#### User story

As a hospital clinical imaging team administrator, I want PIXL to only query for DICOM instances that it doesn't already have, rather pulling an entire study again if it has failed mid-transfer

### Testing

set up unit test with partial data in sent to orthanc raw, full data in VNA and ensure that its reprocessed. Perhaps putting a spy on the method to query for a single instance and ensure that its only the untransferred instances

### Documentation

*No response*

### Dependencies

*No response*

### Details and Comments

Did some investigation with @HChughtai and it looks like its reasonably nice to find the instance UIDs. Suggested flow would be something like this

* Check if study exists in orthanc raw
  * if it doesn't exist, query at the study level as we currently do
  * if exists,
    * if the number of instance UIDs matches between VNA and orthanc raw, if the pixl project tag matches then send it to orthanc anon and return early
    * if they don't have the same number then build a set of all of the instance UIDs in the VNA. Compare this with a set of instance UIDs in orthanc raw.
      * download any missing instances,
  * update the pixl project tag value as we currently do
  * send to orthanc anon directly (and remove our sending on DICOM stable in the `pixl` plugin in orthanc raw

I *think* pyorthanc can do async requests so could be a good time to move to that instead of rolling our own. Here's an example that we did using curl using the values

```
curl --request POST \
  --noproxy "*" \
  -u "pixl:${PIXL_PASSWORD}" \
  --url http://localhost:9006/modalities/UCVNAQR/query \
  --data '{
            "Level" : "Study",
            "Query" : {
              "PatientID" : "${PATIENT_ID}",
              "AccessionNumber": "${ACCESSION_NUMBER}"
            }
          }'
```

> {
> "ID" : "${STUDY_QUERY_ID}",
> "Path" : "/queries/${STUDY_QUERY_ID}"
> }(

studies found

```
curl --noproxy "*" \
  -u "pixl:${PIXL_PASSWORD}" \
  --url http://localhost:9006/queries/${STUDY_QUERY_ID}/answers
```

> \[
> "0",
> "1"
> \](

get the instances for a study, would need to iterate over studies if there are multiple (which does happen by mrn and accession number)

```
curl --request POST \
 --noproxy "*" \
  -u "pixl:${PIXL_PASSWORD}" \
  --url http://localhost:9006/queries/${STUDY_QUERY_ID}/answers/0/query-instances \
  --data '{
            "Query" : {
            },
            "Timeout": 0
          }'
```

> {
> "ID" : "${INSTANCES_QUERY_ID}",
> "Path" : "/queries/${INSTANCES_QUERY_ID}"
> }(

total number of instances, could check if the number of instances is the same between the two servers. if not then compare each SOPInstanceUID

```
curl --noproxy "*" \
  -u "pixl:${PIXL_PASSWORD}" \
  --url http://localhost:9006/queries/${INSTANCES_QUERY_ID}/answers
```

> \[
> "0",
> ...
> "652"
> \]

Get the instance uid (SOPInstanceUID)

```
curl --noproxy "*" \
  -u "pixl:${PIXL_PASSWORD}" \
  --url http://localhost:9006/queries/${INSTANCES_QUERY_ID}/answers/0/content
```

> { "0002,0003" : { "Name" : "MediaStorageSOPInstanceUID", "Type" : "String", "Value" : "REDACTED" }, "0008,0005" : { "Name" : "SpecificCharacterSet", "Type" : "String", "Value" : "ISO_IR 100" }, "0008,0018" : { "Name" : "SOPInstanceUID", "Type" : "String", "Value" : "REDACTED" }, "0008,0050" : { "Name" : "AccessionNumber", "Type" : "String", "Value" : "REDACTED" }, "0008,0052" : { "Name" : "QueryRetrieveLevel", "Type" : "String", "Value" : "IMAGE" }, "0008,0054" : { "Name" : "RetrieveAETitle", "Type" : "String", "Value" : "REDACTED" }, "0010,0020" : { "Name" : "PatientID", "Type" : "String", "Value" : "REDACTED" }, "0020,000d" : { "Name" : "StudyInstanceUID", "Type" : "String", "Value" : "REDACTED" }, "0020,000e" : { "Name" : "SeriesInstanceUID", "Type" : "String", "Value" : "REDACTED" }

p-j-smith:
## Description
Fixes #486 

When querying Orthanc Raw for an existing study, if the study exists but instances are missing, retrieve the missing instances rather than deleting and retrieving the study again.

- Update `_process_message` to retrieve missing instances from the archives:
  - if a study exists in Orthanc Raw:
    - if multiple studies exist in Orthanc Raw, only keep the last updated (previously the study with the longest datetime was kept, rather than the most recent datetime)
    - query the VNA / PACS to determine if any instances are missing
    - retrieve any missing instances
  - if it doesn't exist in Orthanc Raw:
    - query the VNA / PACS for the study
    - retrieve the study from the archives
  - update the project name tag if it's not correct in Orthanc Raw
  - send the study to Orthan Anon. The functionality to auto-route from Orthanc Raw to Orthanc Anon has been removed
- add a test that retrieves a study with two instances into Orthanc Raw. One instance is deleted and the same message is processed. Check the logs to ensure only the missing instances was retrieved
- update `test_processing.py` to use fixtures rather than module level variables

Depends on #495 

## Type of change

### Suggested Checklist

- [ ] I have requested PR review to `UCLH-Foundtry/arc-dev`
- [ ] I have adddressed and marked as resolved all the review comments in my PR.


p-j-smith:
The tests are failing but I _think_ they should pass once #495 is merged

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -212,6 +248,6 @@ async def raise_if_pending_jobs(self) -> None:
     def aet(self) -> str:
         return str(config("ORTHANC_RAW_AE_TITLE"))
 
-    async def send_existing_study_to_anon(self, resource_id: str) -> Any:
+    async def send_study_to_anon(self, resource_id: str) -> Any:
```
stefpiatek, line 251: oh actually, we should make this an `"Synchronous": False,` job and wait for the job to complete

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -99,11 +119,12 @@ async def modify_private_tags_by_study(
                 "Force": True,
                 "Keep": ["StudyInstanceUID", "SeriesInstanceUID", "SOPInstanceUID"],
```
stefpiatek, line 120: We should also use the "Timeout" here as the dicom timeout, hopefully that works

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -99,11 +119,12 @@ async def modify_private_tags_by_study(
...
             },
+            timeout=self.http_timeout,
         )
         logger.debug("Modify studies Job: {}", response)
         job_id = str(response["ID"])
         try:
-            await self.wait_for_job_success_or_raise(job_id, "modify", timeout=timeout)
+            await self.wait_for_job_success_or_raise(job_id, "modify", timeout=self.http_timeout)
```
stefpiatek, line 127: this one should be the DICOM timeout, as its for the entire data to be returned

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -113,12 +134,27 @@ async def retrieve_from_remote(self, query_id: str) -> str:
         response = await self._post(
             f"/queries/{query_id}/retrieve",
             data={"TargetAet": self.aet, "Synchronous": False},
```
stefpiatek, line 136: can also set the "Timeout" in the posted JSON

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -113,12 +134,27 @@ async def retrieve_from_remote(self, query_id: str) -> str:
...
+        self, modality: str, missing_instances: list[dict[str, str]]
+    ) -> str:
+        """Retieve missing instances from remote modality in a single c-move query."""
+        response = await self._post(
+            f"/modalities/{modality}/move",
+            data={
+                "Level": "Instance",
+                "TargetAet": self.aet,
+                "Synchronous": False,
+                "Resources": missing_instances,
```
stefpiatek, line 151: want to use the JSON Timeout here for the entire cmove job timing out

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -113,12 +134,27 @@ async def retrieve_from_remote(self, query_id: str) -> str:
...
+            },
+            timeout=self.dicom_timeout,
```
stefpiatek, line 153: this one should be the http timeout as its just waiting for the job id from the POST

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -212,6 +256,10 @@ async def raise_if_pending_jobs(self) -> None:
...
         return str(config("ORTHANC_RAW_AE_TITLE"))
 
-    async def send_existing_study_to_anon(self, resource_id: str) -> Any:
+    async def send_study_to_anon(self, resource_id: str) -> Any:
         """Send study to orthanc anon."""
-        return await self._post("/send-to-anon", data={"ResourceId": resource_id})
+        return await self._post(
+            "/send-to-anon",
+            data={"ResourceId": resource_id, "Synchronous": False},
+            timeout=self.dicom_timeout,
```
stefpiatek, line 264: changing of timeouts, but we will then also need to monitor the job to make sure it succeeds, like we do with the c-move of DICOM

```suggestion
            data={"ResourceId": resource_id, "Synchronous": False, "Timeout": self.dicom_timeout},
            timeout=self.http_timeout,
```

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -113,12 +134,27 @@ async def retrieve_from_remote(self, query_id: str) -> str:
         response = await self._post(
             f"/queries/{query_id}/retrieve",
             data={"TargetAet": self.aet, "Synchronous": False},
+            timeout=self.dicom_timeout,
```
stefpiatek, line 137: and then this one should be the http timeout

stefpiatek (commented):
Had a quick look and some suggestions on the timeout, I think I got more clear as I went on but let me know!

p-j-smith:
> Had a quick look and some suggestions on the timeout, I think I got more clear as I went on but let me know!

Ah thanks for the explanations! I now get the difference between "Timeout" in the json and the `timeout` argument

orthanc/orthanc-raw/plugin/pixl.py
```diff
@@ -51,36 +49,6 @@
 logger.warning("Running logging at level {}", logging_level)
```
stefpiatek, line 51: I wonder if we should try this (and registering the the callback again below) instead of sending directly to anon in `process_message`. Could be that it needs to be stable to be kept?
```suggestion
def OnChange(changeType, level, resourceId):  # noqa: ARG001
    """
    # Taken from:
    # https://book.orthanc-server.com/plugins/python.html#auto-routing-studies
    This routes any stable study to a modality named PIXL-Anon if
    should_auto_route returns true
    """
    if changeType != orthanc.ChangeType.STABLE_STUDY:
        return

    # imaging-api didn't get to updating pixl project name, so incomplete
    if _get_project_name_from_study_id(resourceId) == DICOM_TAG_PROJECT_NAME.PLACEHOLDER_VALUE:
        logger.warning("Study {} has not been set with a pixl project name, not sending" resourceId)
        return

    if should_send_to_anon():
        logger.debug("Sending study: {}", resourceId)
        # Although this can throw, since we have nowhere to report errors
        # back to (e.g. an HTTP client), don't try to handle anything here.
        # The client will have to detect that it hasn't happened and retry.
        orthanc_anon_store_study(resourceId)
```

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -148,11 +181,15 @@ async def job_state(self, job_id: str) -> Any:
...
+            session.get(
+                f"{self._url}{path}",
+                auth=self._auth,
+                timeout=self.http_timeout,
+            ) as response,
         ):
             return await _deserialise(response)
 
-    async def _post(self, path: str, data: dict, timeout: Optional[float] = None) -> Any:
+    async def _post(self, path: str, data: dict, timeout: int) -> Any:
```
stefpiatek, line 192: I wonder if all posts should just use `self.http_timeout` and not take this as an argument?

pixl_imaging/src/pixl_imaging/_orthanc.py
```diff
@@ -212,6 +253,11 @@ async def raise_if_pending_jobs(self) -> None:
...
-    async def send_existing_study_to_anon(self, resource_id: str) -> Any:
+    async def send_study_to_anon(self, resource_id: str) -> str:
         """Send study to orthanc anon."""
-        return await self._post("/send-to-anon", data={"ResourceId": resource_id})
+        reponse = await self._post(
+            "/send-to-anon",
+            data={"ResourceId": resource_id, "Synchronous": False, "Timeout": self.dicom_timeout},
+            timeout=self.http_timeout,
+        )
+        return str(reponse["ID"])
```
stefpiatek, line 263: Ah this is a custom endpoint, so we'd need to update the `pixl.py` plugin to use the timeout value when sending and do it async. Worth trying out to see if the OnChange callback only sending if the study is stable and the project name is set is required for the study to be kept in orthanc raw first. If it is then go back to using that I guess, and then updating the send to be asynchronous

stefpiatek (commented):
Looking good, added in some suggestions and I think there's two things left:

- blocking: it appears that studies that are successfully exported aren't kept in orthanc raw now
- would be nice to solve: orthanc raw becomes unresponsive after 4-8 hours
  - hunch here is the the JOB-INS counter keeps on increasing and that causes it. Not sure why it only ever increases and never decreases. 
  - Setting the number of threads for DICOM modification to `5` helped, could try setting back to the default of `1`?

pixl_imaging/src/pixl_imaging/_orthanc.py
p-j-smith: ah yeah that makes more sense

orthanc/orthanc-raw/plugin/pixl.py
```diff
@@ -169,27 +136,19 @@ def log_and_return_http(
...
-            output,
-            200,
-            "Auto-routing is not enabled",
-            f"Auto-routing is not enabled, dropping request {request}",
-        )
-        return
     try:
         body = json.loads(request["body"])
-        resource_id = body["ResourceId"]
+        body["Resources"]
```
p-j-smith, line 141: according to the [api docs for c-store](https://orthanc.uclouvain.be/api/#tag/Networking/paths/~1modalities~1{id}~1store/post) we need to specify a list of resources now that we're sending json

orthanc/orthanc-raw/plugin/pixl.py
p-j-smith: I tried adding this back in but ran into the same errors. Turns out the issue was when `study_uid` is an empty string, the function `Orthanc.query_local` would still query by UID and return all studies. Then all but the most recent study was deleted. It's fixed now by checking that `study_uid` is not empty before using it to query studies

pixl_imaging/tests/docker-compose.yml
```diff
@@ -85,7 +85,6 @@ services:
             ORTHANC_ANON_AE_TITLE: "unused"
             ORTHANC_ANON_DICOM_PORT: "4242"
```
stefpiatek, line 86: mentioned it before but might be worth setting restart to always

pixl_imaging/src/pixl_imaging/_processing.py
```diff
@@ -44,108 +44,161 @@ async def process_message(message: Message) -> None:
...
+        )
 
+    logger.debug("Local instances for study: {}", resource)
+
+    if config("ORTHANC_AUTOROUTE_RAW_TO_ANON", default=False, cast=bool):
+        job_id = await orthanc_raw.send_study_to_anon(resource_id=resource["ID"])
+        await orthanc_raw.wait_for_job_success_or_raise(
+            job_id, "c-store", timeout=orthanc_raw.dicom_timeout
+        )
+        return
```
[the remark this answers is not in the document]
stefpiatek: seems like a good shout

stefpiatek (approved):
Looking nice, thanks for digging into it deeply. Some suggestions and hopefully orthanc raw plays nicely now

orthanc/orthanc-raw/plugin/pixl.py
```diff
@@ -169,27 +136,19 @@ def log_and_return_http(
 def SendResourceToAnon(output, uri, **request):  # noqa: ARG001
```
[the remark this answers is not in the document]
p-j-smith: Yeah definitely, that's a good call

orthanc/orthanc-raw/plugin/pixl.py
```diff
@@ -51,41 +49,18 @@
...
-
-def orthanc_anon_store_study(resource_id):
+def orthanc_anon_store_study(body):
     """Call the API to send the specified resource (study) to the orthanc anon server."""
     # RestApiPost raises an orthanc.OrthancException if it fails
-    orthanc.RestApiPost("/modalities/PIXL-Anon/store", resource_id)
-    orthanc.LogInfo(f"Successfully sent study to anon modality: {resource_id}")
+    body["Asynchronous"] = True
+    response = orthanc.RestApiPost(
+        "/modalities/PIXL-Anon/store",
```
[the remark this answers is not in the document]
p-j-smith: ooo yeah that's probably a good idea

pixl_imaging/tests/docker-compose.yml
p-j-smith: ah yep, good point

pixl_imaging/src/pixl_imaging/_orthanc.py
p-j-smith: marking as resolved as we've moved `send_study_to_anon` out of the plugin and into `pixl_imaging._processing`

p-j-smith:
> Looking good, added in some suggestions and I think there's two things left:
> - blocking: it appears that studies that are successfully exported aren't kept in orthanc raw now
> - would be nice to solve: orthanc raw becomes unresponsive after 4-8 hours

Thanks for the helpful reviews! I think both of those issues are fixed now. The test instance has been up and running for nearly 24 hours and is the rate of message processing is pretty stable:
image

PR #510, opened 2024-10-08
Check secondary hours before logging processing

stefpiatek:
That way don't have INFO level logs of processing messages from secondary queue before 8pm


p-j-smith (approved):
ah that's a good call!

PR #515, opened 2024-10-17
Reset start time if pending status

stefpiatek:
If a job is pending for a long time then it can timeout currently, which shouldn't happen


p-j-smith (approved):
ah good catch, is that why lots of jobs were failing?

stefpiatek:
Yeah it could be part of it, though I we only get very long pending if there's other issues happening like we can't find the `Image` entity in the pixl database

Issue #508, opened 2024-10-02
Create human readable name for session/experiment


stefpiatek:
Work out a sensible, unique (per project) name for the experiment or session

p-j-smith:
After discussing with @HChughtai, the easiest approach for now is to continue using the `StudyInstanceUID` to set the experiment label. We can then rename experiments in XNAT using a custom script for each project.

Going forward, we could implement the same [label generation](https://bitbucket.org/xnatdev/xnat-desktop-client/src/master/services/experiment_label.js) as the XNAT Desktop Client: 

```python
f"{patient_id}_{modality}_{count}"
```

Note:
- we probably can't base this off something in the PIXL DB as users may upload data outside of PIXL and we want to avoid clashes with experiment labels. So we will need to query XNAT for existing experiment labels for the project / subject
- if there are multiple modalities in the DICOM study, we will create a label that includes all modalities

stefpiatek:
I think this has been closed, but reopen if I'm wrong 🙏🏻

PR #525, opened 2024-11-25
Add configuration for IoN Neuro DB project

stefpiatek:
Creating default configuration for IoN Neuro DB project. 

We override sequentually so the first file to be read is the
- [projects/configs/tag-operations/base.yaml](https://github.com/SAFEHR-data/PIXL/blob/main/projects/configs/tag-operations/base.yaml)
- then [projects/configs/tag-operations/mri.yaml](https://github.com/SAFEHR-data/PIXL/blob/main/projects/configs/tag-operations/mri.yaml)
- and specific to this project: [projects/configs/tag-operations/ion-neuro-db.yaml](https://github.com/SAFEHR-data/PIXL/blob/main/projects/configs/tag-operations/ion-neuro-db.yaml)
- Then for specific manufacturers: [projects/configs/tag-operations/manufacturer-overrides/mri.yaml](https://github.com/SAFEHR-data/PIXL/blob/main/projects/configs/tag-operations/manufacturer-overrides/mri.yaml)

### DIFF
--- a/projects/configs/ion-neuro-db.yaml
+++ b/projects/configs/ion-neuro-db.yaml
@@ -1,4 +1,4 @@
-#  Copyright (c) 2022 University College London Hospitals NHS Foundation Trust
+#  Copyright (c) 2024 University College London Hospitals NHS Foundation Trust
 #
 #  Licensed under the Apache License, Version 2.0 (the "License");
 #  you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


p-j-smith (approved):
looks good! Does https://github.com/SAFEHR-data/PIXL/issues/508 need to be addressed before exporting data to the XNAT?

stefpiatek:
> looks good! Does #508 need to be addressed before exporting data to the XNAT?

Yeah I think practially we won't actually export any data as part of this tranche of work, but setting up the configuration so we can test out the anonymisation within UCLH network. I guess I could temporarily just set both exports to `none`, but also there's no authentication set up to allow export so can't accidentaly do that

PR #528, opened 2024-11-26
Tweak logging for anonymisation

stefpiatek:


p-j-smith (approved):
very nice - will be good to reduce the noise from the anonymisation

PR #532, opened 2024-12-03
Fix timeout issues when querying instances of a local study in Orthanc

### LINKED ISSUE #531: Querying instances of local studies in Orthanc Raw is slow and times out
### What happened?

The query to get a list of instances in a local study in Orthanc Raw is sometimes very slow (10+ minutes) and leads to `Timeout` errors in the imaging api.

### Relevant log output (optional)

_No response_

### What did you expect? (optional)

I would expect the call to `/studies/id/instances` should return within the HTTP timeout limit.

I've tried:

- calling different endpoints (e.g. `/studies/series` to get a list of series in the study and then `/series/instances` to get the instances in a series, but the call to `/series/instances` is very slow if there are a large number of instances in the series)
- specifying only the `SOPInstanceUID` tag should be returned in the call to `/study/id/instances` (although all MainDicomTags are still returned)
- making the call to `/studies/id/instances` synchronously

but none of these to solve the issue. We should perhaps use the DICOM timeout in this api call so we at least don't run into the timeout issues.

p-j-smith:
## Description
Fixes #531

- use the DICOM timeout when getting info on the instances in a study
- when determining if there are any instances missing, first get statistics of the study and compare the number of local and remote instances (if all instances are present locally there's no need to get the list of `SOPInstanceUIDs`)

## Type of change

### Suggested Checklist

### DIFF
--- a/pixl_imaging/src/pixl_imaging/_orthanc.py
+++ b/pixl_imaging/src/pixl_imaging/_orthanc.py
@@ -64,6 +64,10 @@ class Orthanc:
         """Query local Orthanc instance for study."""
         return await self._get(f"/studies/{study_id}")
 
+    async def get_local_study_statistics(self, study_id: str) -> Any:
+        """Query local Orthanc instance for study statistics."""
+        return await self._get(f"/studies/{study_id}/statistics")
+
     async def get_local_study_instances(self, study_id: str) -> Any:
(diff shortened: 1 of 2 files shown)


stefpiatek (approved):
🕐

PR #572, opened 2024-12-12
Add configuration for DESPIAD project

### LINKED ISSUE #530: PIXL pipeline for CT data
Not a modality we can manage yet; there will be a clear need; should do when we have a good clinical use case

### LINKED ISSUE #571: Add config for DESPIAD

- [x] add a project config for DESPIAD
- [x] add anonymisation config for PET data
- [x] add anonymisation config for CT data #530 
- [ ] ~~add a sensible default session labelling scheme for XNAT~~ not required, see https://github.com/SAFEHR-data/PIXL/issues/508#issuecomment-2548379677

### Testing

- [x] Check with Dave / Will that the de-identified data is suitable for their needs

### Documentation

_No response_

### Dependencies

- [x] #530 

### Details and Comments

_No response_

p-j-smith:
## Description
Closes #571 
Closes #530 

- add project config for DESPIAD
- add tag operation files for CT and PET
- use base, CT, then PET tag operation files for anonymisation
- ~~ensure all resources sharing a `StudyInstanceUID` in Orthanc Raw are exported (they may exist as separate resources in Orthanc Raw if they have different `StudyID`s, but are combined into a single resource in Orthanc Anon as the `StudyID` tag is deleted)~~ This has already been fixed in #582 

## Type of change

### Suggested Checklist


projects/configs/despiad.yaml
```diff
@@ -0,0 +1,32 @@
...
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+project:
+    name: "despiad"
+    modalities:
+      - "CT"
```
stefpiatek, line 18: with CT AFAIK there's some extra fun parts which have burned in data. I think to avoid this we may want to have:


- Minimum instances in a series. At the project level can configure minimum instances in series so that we can filter out a single screenshot that may well still have the `CT` modality
- Series filter at a manufacturer level in the project conifg. This is because certain series numbers can always have burned in data - Worth asking John Watts about an example of this
- Related, but to have an allowed manufacturer regex list in the config. So that you filter out manufacturers which you don't know you can handle - in case there's more burned in data

p-j-smith: oh fun! Yeah they sound like good ideas. Have you ever used something like [deid](https://github.com/pydicom/deid) or [presidio](https://github.com/microsoft/presidio) to remove burned in data?

p-j-smith: I've add these three new options to the project config. The only one I'm not sure about is the series number filter at the manufacturer level. Currently, I've added it at the project level, but do you mean we should be able to specify that the filter is only applied to DICOM from a specified manufacturer? Would something like this make sense:
```
manufacturer_filters:
  - <manufacturer>:
    - series_number: [<series number(s) to filter>]

```

stefpiatek: > oh fun! Yeah they sound like good ideas. Have you ever used something like [deid](https://github.com/pydicom/deid) or [presidio](https://github.com/microsoft/presidio) to remove burned in data?

oh whoops I missed this. Nah haven't tried using them but that could be useful for areas that we know we want to redact specifically. Would be concerned about running them wholesale on all images

stefpiatek: Yeah I think have the allowed manufacturers in the filter perhaps. so like this maybe?
```yaml
manufacturer_filters:
  - regex: ^siemens
    exclude_series_numbers: [<series number(s) to filter>]
  - regex: ^philips
    exclude_series_numbers: [<series number(s) to filter>]
```

p-j-smith: ooo yeah that's nicer. Although I may misunderstand this:

```python
  - regex: ^siemens # defaults to blank, so all manufacturers would be skipped.
```
If it defaults to blank, then wouldn't all manufacturers be included (as the regex is for allowed manufactures and a blank string matches everything")

stefpiatek: oh does a blank string match everything? I would have thought `*` would be the only way to match all. If so then could do default of a `^$`

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -339,6 +340,15 @@ def _anonymise_study_instances(
...
             logger.debug("Reading file {}", file)
             dataset = dcmread(file)
+
+            if dataset.SeriesInstanceUID in series_to_skip:
+                logger.warning(
+                    "Skipping series {} for study {} due to too few instances",
+                    dataset.SeriesInstanceUID,
+                    study_info,
+                )
+                continue
```
stefpiatek, line 350: Oh we should also add this to the `skipped_instance_counts` and change this to debug to follow what we've done for other ways of skipping

projects/configs/despiad.yaml
p-j-smith: yeah I just checked and `bool(re.search("", "123"))` returns `True`. I was thinking of using a super explicit default like `no manufacturers allowed ^`, just to make it obvious to anyone reading the code. Also, not sure if it's possible but if there's any chance the manufacturer could be an empty string, then `bool(re.search("^$", ""))` would be `True`

p-j-smith: oh fun! I think that's due to this line:
https://github.com/SAFEHR-data/PIXL/blob/dca65ca12327c232434188e9967448247c0edab5/pixl_dcmd/src/pixl_dcmd/main.py#L140

What's the best way to handle these datasets? Should we raise a `PixlDiscardError`?

p-j-smith: For now I'm raising a `PixlSkipInstanceError` if `Manufacturer` is `None`

stefpiatek: Yeah I think let's go for instance, can always scale up to discard later

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -376,6 +386,33 @@ def _anonymise_study_instances(
     return anonymised_instances_bytes, anonymised_study_uid
 
 
+def get_series_to_skip(zipped_study: ZipFile, min_instances: int) -> set[str]:
```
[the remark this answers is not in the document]
p-j-smith: ah yeah, good point, I've moved it now

orthanc/orthanc-anon/plugin/pixl.py
p-j-smith: ah yes! I've added the counts

pixl_dcmd/tests/test_main.py
```diff
@@ -194,7 +194,9 @@ def ids_for_parameterised_test(val: pathlib.Path) -> str:
 
 
 @pytest.mark.parametrize(
-    ("yaml_file"), PROJECT_CONFIGS_DIR.glob("*.yaml"), ids=ids_for_parameterised_test
+    ("yaml_file"),
+    PROJECT_CONFIGS_DIR.glob("test-*.yaml"),
```
p-j-smith, line 198: no longer testing anonymisation using the non-test project configs. If we want to test all project configs, we need to add:
```yaml
allowed_manufacturers:
  - regex: "^company"
    exclude_series_numbers: []
```
to all configs, unless there's a nicer way to do it?

p-j-smith (commented):
do we know the `allowed_manufactures` we should add for each project?

pixl_dcmd/tests/test_main.py
stefpiatek: I think for the existing projects we can just allow all manufacturers, as that's what was implicitly done

p-j-smith: ah yeah that's fair, and for DESPIAD shall we limit to known manufacturers? And yeah we have a [separate test](https://github.com/SAFEHR-data/PIXL/blob/4afca8acd700f91e7b135da2ec71b394c4e859b3/pixl_core/tests/project_config/test_project_config.py#L91-L97) to check configs are valid

p-j-smith: I've allowed all manufacturers for the existing configs, and added them back to the `test_anonymise_and_validate_dicom` tests

projects/configs/despiad.yaml
```diff
@@ -0,0 +1,39 @@
...
+    base:
+        - "base.yaml"
+        - "ct.yaml"
+        - "pet.yaml"
+        - "despiad.yaml"
+    manufacturer_overrides: []
+
+allowed_manufacturers:
+  - regex: ".*"
+    exclude_series_numbers: []
```
stefpiatek, line 31: I think we should go through and find out all of the manufacturers before we allow the all through, as CT can have burned in data

projects/configs/despiad.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+min_instances_per_series: 1
```
stefpiatek, line 33: Are we expecting any single instances? If not I'd suggest `2` so that screenshots aren't uploaded

template_config.yaml
```diff
@@ -22,6 +22,18 @@ tag_operation_files: # DICOM tag anonymisation operations
         - "base-tag-operations.yaml" # Base schema
     manufacturer_overrides: none # Manufactuer-dependendent overrides
 
+allowed_manufacturers:
+  - regex: ".*"  # allow all manufacturers
+    exclude_series_numbers: []
+
+min_instances_per_series: 1
```
stefpiatek, line 29:
```suggestion
min_instances_per_series: 2
```

template_config.yaml
```diff
@@ -22,6 +22,18 @@ tag_operation_files: # DICOM tag anonymisation operations
         - "base-tag-operations.yaml" # Base schema
     manufacturer_overrides: none # Manufactuer-dependendent overrides
 
+allowed_manufacturers:
+  - regex: ".*"  # allow all manufacturers
+    exclude_series_numbers: []
```
stefpiatek, line 27: I think our template shouldn't allow all through, because that opens us up to exporting series which are burned in. But adding in canon's 8000 series to exclude is a good idea

stefpiatek (approved):
Looking good, but I think we need to be careful about despiad series to filter out. At least checking an example from each manufacturer for each modality to ensure that we're not exporting burned in PID

projects/configs/despiad.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+allowed_manufacturers:
+  - regex: "^ge medical systems"
+    exclude_series_numbers:
+        - "1200"
+        - "1201"
+        - "1202"
+        - "1203"
+        - "1301"
+        - "1501"
```
stefpiatek, line 37: Imagine these get converted into numbers, or are the series in pydicom strings?

template_config.yaml
```diff
@@ -20,7 +20,23 @@ project:
...
-    manufacturer_overrides: none # Manufactuer-dependendent overrides
+    manufacturer_overrides: [] # Manufactuer-dependendent overrides
+
+allowed_manufacturers:
+  - regex: "^example-manufacturer"
+    exclude_series_numbers: []
+  # For DICOM generated by Canon, Series "8000" always has bunred-in data
+  # so we always exlcude this series
+  - regex: "^canon"
+    exclude_series_numbers: ["8000"]
```
stefpiatek, line 31: I guess we could also add in the GE excluded series in the template as well

projects/configs/despiad.yaml
p-j-smith: they have type `pydicom.valuerep.IS`, which can be compared against either strings or ints, but you're right having them as ints makes more sense I think

template_config.yaml
p-j-smith: oh yeah good point. I've also added `mip` to the excluded series description in the template (or should I remove it?)

stefpiatek: Yeah I think its a good idea, could add a comment that for PET studies this has PID

projects/configs/despiad.yaml
p-j-smith: good point, we're now limiting to `^ge medical systems`

README.md
```diff
@@ -147,6 +147,13 @@ The configuration file defines:
 
 - Project name: the `<project-slug>` name of the Project
 - The DICOM dataset modalities to retain (e.g. `["DX", "CR"]` for X-Ray studies)
+- The minimum number of instances required by a series (defaults to 1). May be set higher than 1 to filter out
```
p-j-smith, line 150: should `min_instances_per_series` default to 2?

projects/configs/tag-operations/pet.yaml
```diff
@@ -0,0 +1,222 @@
...
+  element: 0x0022
+  op: replace
+- name: Series Time
+  group: 0x0008
+  element: 0x0031
+  op: keep
+- name: Acquisition Time
+  group: 0x0008
+  element: 0x0032
+  op: keep
```
p-j-smith, line 30: Should we be replacing the times by default? Dave [suggested](https://github.com/SAFEHR-data/PIXL/pull/592#issue-2779926957) we default to keeping them:

> More information about the acquisition (reconstrution, randoms correction, etc) is also useful to ensure study protocol compliance in single and multi-centre studies. Acquisition time is a Level 1 attribute for PET series. And I think it may be worth keeping time to ensure uptake time, especially as in the PET base module all of the dates are either replace or wiped out.

p-j-smith (commented):
I think it's nearly ready to merge, just a couple of questions about what defaults we want

README.md
stefpiatek: Yeah I think so

projects/configs/tag-operations/pet.yaml
stefpiatek: I think we're probably okay but might be worth dropping Sophie an email to get her take on it from an IG perspective

p-j-smith: Yeah good idea. Sophie confirmed it's okay to keep the times

PR #573, opened 2024-12-12
Renames file to prevent pytest conflict

tomaroberts:
Very minor PR which renames file to prevent pytest from throwing an error when running all tests together.


stefpiatek (approved):
Lovely, thanks Tom

PR #574, opened 2024-12-12
Start documenting design choices

stefpiatek:
Actually documenting design choices. Hopefully a reasonable summary of what the options considered were, at least in the last year or so.

What else have I missed?

Reckon there'll be a lot so up for other suggestions of decisions that we've made that are worth noting


milanmlft (approved):
Looking really nice 👌 
Wonder if it's worth adding a record for the export API? From memory I seem to recall there was some back and forth on how to approach this as well, as we  needed something flexible that can handle multiple destinations on a per-project basis.

stefpiatek:
> Looking really nice 👌 Wonder if it's worth adding a record for the export API? From memory I seem to recall there was some back and forth on how to approach this as well, as we needed something flexible that can handle multiple destinations on a per-project basis.

Oh yeah that's a fair point. Let me track down what we said about it

tomaroberts:
@stefpiatek

I have had a read through this and it's clearly useful documentation and I found the context helpful. I can't contribute hugely as of yet, but please don't let me hold you up if you want to merge. Happy to revisit once I'm more familiar with the project.

PR #579, opened 2024-12-16
Switch to `uv` python tool

jeremyestein:
## Description
Switch to using the `uv` Python package manager in all places: docker builds, GHA, and dev instructions.

Known TODOs:
 - [x] documentation still refers to `pip`
 - [x] explicitly copying all pyproject.toml files in the Dockerfiles is a bit ugly, but can't find a better way
 - [x] The base branch needs some work (see PR comment)
 - [x] `uv sync --all-extras --all-packages` is currently needed to get pytest. I'd expect just extras to be needed. Do I need to add extras section to root pyproject and link to workspace extras?
 - [x] Do we need a mechanism to ensure that changes to `pyproject.toml` files are reflected in the `uv.lock` file so it can be committed too?
 - [x] Why do we need to specify `setuptools` in pixl_dcmd just for the sake of pydicom? Shouldn't it be responsible for its own dependencies?
 - [x] Do we need to make use of dependency groups?

## Type of change

- [x] This change requires a documentation update

### Suggested Checklist

- [x] I have passed on my local host device. (see further details at the [CONTRIBUTING](https://github.com/SAFEHR-data/PIXL/blob/main/CONTRIBUTING.md#local-testing) document)
- [x] Make sure your branch is up-to-date with main branch. See [CONTRIBUTING](https://github.com/SAFEHR-data/PIXL/blob/main/CONTRIBUTING.md) for a general example to syncronise your branch with the `main` branch.
- [x] I have requested review to this PR.
- [x] I have addressed and marked as resolved all the review comments in my PR.


.github/workflows/main.yml
```diff
@@ -56,22 +56,23 @@ jobs:
...
+                  enable-cache: true
+
 
             - name: Install Python dependencies
               run: |
-                  python -m pip install --upgrade pip
-                  pip install -e pytest-pixl/
-                  pip install -e pixl_core/[test]
-                  pip install -e ${{ matrix.package_dir }}/[test]
+                  uv sync --all-extras --all-packages
```
p-j-smith, line 69: to avoid needing the `--all-extras` and `--all-packages` flags, you can add a `dev` [dependency group](https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies) in the top-level `pyproject.toml`(you can move dev dependencies from all packages into the top-level toml). The dev dependencies are installed by default when you do `uv sync`.

Also, it's probably best to do `uv sync --frozen` to avoid updating the `uv.lock` file before running tests

pixl_dcmd/pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
     "pydicom-data",
     "python-decouple==3.8",
     "requests==2.32.3",
+    "setuptools", # pydicom needs this
```
p-j-smith, line 19: I deleted this dependency and tried the following:
```python
uv python install 3.11
uv venv --python 3.11
source .venv/bin/activate
uv sync
```
and it seemed to install okay. What was the error you ran into?

p-j-smith (commented):
this is looking really good - will be nice to switch to uv. Just a couple of comments at the minute, plus:

>  Do we need a mechanism to ensure that changes to pyproject.toml files are reflected in the uv.lock file so it can be committed too?

[renovate _should_ update both](https://docs.astral.sh/uv/guides/integration/dependency-bots/#renovate) the `pyproject.toml` and `uv.lock` files at the same time

pixl_dcmd/pyproject.toml
jeremyestein: I think this will give an error in the system tests when orthanc tries to run or load plugins.

jeremyestein:
> this is looking really good - will be nice to switch to uv. Just a couple of comments at the minute, plus:
> 
> > Do we need a mechanism to ensure that changes to pyproject.toml files are reflected in the uv.lock file so it can be committed too?
> 
> [renovate _should_ update both](https://docs.astral.sh/uv/guides/integration/dependency-bots/#renovate) the `pyproject.toml` and `uv.lock` files at the same time

(To record our earlier discussion) I was talking was about developer-initiated changes to `pyproject.toml`. This has been solved by adding a pre-commit check that runs `uv lock` which brings the lock file up to date.

.github/workflows/main.yml
jeremyestein: Should have fixed this now. It looked like we weren't making a huge distinction between dev and test optional packages, so I merged them all into a top-level `dev` dependency group.

Re the `--frozen` option, since we now have the pre-commit check making sure that the lock file has been updated, I think I'm going to run the tests without this option, so that in the event that you forget to update the lock file before pushing, you get the test results for the lock file you were *supposed* to have, which is I think more useful than using the old version. Unless you object.

jeremyestein:
I have changed all the `pip install` commands in the docs to basically a `uv sync`. I wasn't always sure of the intent of the docs that I changed, but fixing that is a separate task ( #591 ).

pixl_dcmd/pyproject.toml
jeremyestein: I tested it and the system test works fine without it. Not sure why it fixed something at the time!

jeremyestein: Ah sorry I misremembered, it was the pixl_dcmd test that fails without it.

docker/pixl-python/Dockerfile
```diff
@@ -16,7 +16,8 @@ SHELL ["/bin/bash", "-o", "pipefail", "-e", "-u", "-x", "-c"]
 
 ARG TEST="false"
 
-RUN export DEBIAN_FRONTEND=noninteractive && \
+RUN <<EOF
+    export DEBIAN_FRONTEND=noninteractive && \
```
stefpiatek, line 20: if we're doing EOF, do we need the trailing `\`?

docs/setup/developer.md
```diff
@@ -2,6 +2,18 @@
...
+Create the venv (one-off task)
+```
+uv venv
+```
+Enter the venv (for the lifetime of the shell)
+```
+source .venv/bin/activate
+```
+
 ### Using conda
```
stefpiatek, line 17: I'd take out the conda and venv options, and only have `uv` as the only way that we document

pyproject.toml
```diff
@@ -0,0 +1,69 @@
+[project]
+name = "pixl"
+version = "0.2.0rc0"
+description = "Top-level PIXL project"
+readme = "README.md"
+requires-python = ">=3.9"
```
stefpiatek, line 6: We have `requires-python = ">=3.11"` for pixl_dcmd so should be consistent across all our modules

stefpiatek (approved):
Chef's kiss. Some suggestions, happy for you to merge without another review after you've addressed them to your satisfaction

.github/workflows/main.yml
p-j-smith: > I think I'm going to run the tests without this option, so that in the event that you forget to update the lock file before pushing, you get the test results for the lock file you were supposed to have

yeah I think you're right actually, that makes more sense

docs/setup/developer.md
```diff
@@ -2,6 +2,18 @@
 
 ## Setting up `Python` Virtual Environment (VE)
 
+### Using `uv` (stringly recommended)
+Once you have installed `uv`, in the root of your source tree you can:
+
+Create the venv (one-off task)
+```
+uv venv
+```
```
p-j-smith, line 11:
``````suggestion
Once you have installed `uv`, in the root of your source tree you can:

Install the required Python version and create the venv (one-off task)
```
uv python install 3.11
uv venv --python 3.11
```
``````

p-j-smith (approved):
looking great. One minor suggestion around specifying the python version with `uv`. Alternatively, we could add a [`.python-version` file](https://docs.astral.sh/uv/concepts/python-versions/#python-version-files) that `uv` will detect. Happy with whatever you decide

docs/setup/developer.md
jeremyestein: The GAE instructions also had old references to conda. Mostly replaced with link to Slab doc.

docker/pixl-python/Dockerfile
jeremyestein: Ah good point; my IDE adds them back in quite aggresively.

PR #582, opened 2024-12-18
Improve anon logging and anonymise all matching studies at once

stefpiatek:
- Streamline logging and use of stdout to make it study-level
- Anonymise all studies together before exporting - as can have multiple studies matching MRN and accession number. Merging these and sending all together seems reasonable
- Remove unused environmental variables

Maybe could have broken out but also does:

- Set default timeout for export via dicomweb


orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -323,37 +350,56 @@ def _anonymise_study_instances(
...
                 anonymised_study_uid = dataset[0x0020, 0x000D].value
+                dicom_validation_errors |= instance_validation_errors
+
+    if skipped_instance_counts:
+        logger.info(
+            "Project '{}' Study {}, skipped instance counts: {}",
+            project_name,
+            study_uid,
+            dict(skipped_instance_counts),
+        )
```
stefpiatek, line 384: Is this useful here, or should we just report it in the discard error below?

p-j-smith: yeah maybe in the discard error would be good, or make it debug rather than info?

pixl_dcmd/src/pixl_dcmd/_dicom_helpers.py
```diff
@@ -15,23 +15,45 @@
...
+    old_stdout = sys.stdout
+    sys.stdout = StringIO()
+    try:
+        yield
+        output = sys.stdout.getvalue()
+        for line in output.splitlines():
+            _logger.debug(line.strip())
+    finally:
+        sys.stdout = old_stdout
+
```
p-j-smith, line 46: ooo very nice

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -250,24 +283,18 @@ def _import_study_from_raw(study_resource_id: str, study_uid: str, project_name:
                 project_name=project_name,
             )
         except PixlDiscardError as discard:
-            logger.warning("Failed to anonymize study {}: {}", study_uid, discard)
+            logger.warning(
+                "Failed to anonymize project: '{}', study: {}: {}", project_name, study_uid, discard
```
p-j-smith, line 287: ah very nice to include the project in the logs

p-j-smith (approved):
lovely stuff, just one minor suggestion around notifying the export API

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -218,28 +220,59 @@ def ImportStudiesFromRaw(output, uri, **request):  # noqa: ARG001
...
+        for anonymised_study_uid in anonymised_study_uids
+    }
+
+    logger.debug(
+        "Notify export API to retrieve study resources. Original UID {} Anon UID: {}",
+        study_resource_ids,
+        resource_ids,
+    )
+
+    for resource_id in resource_ids:
```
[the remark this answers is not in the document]
stefpiatek: ah so it is already a set unless I'm mistaken, its a set generator expression right?

PR #585, opened 2024-12-27
Make it clear when study already exported

stefpiatek:
Also reworked test setup so there's a bit less duplication, and added in tests


pixl_dcmd/src/pixl_dcmd/_database.py
```diff
@@ -128,27 +128,30 @@ def get_unexported_image(
...
+
+def _query_and_raise_if_exported(
+    pixl_session: Session, clause_list: list[ColumnElement[bool]]
+) -> Image:
+    existing_image: Image = (
+        pixl_session.query(Image).join(Extract).filter(*clause_list).one()
+    )
+    if existing_image.exported_at is not None:
+        msg = "Study already exported"
+        raise exc.NoResultFound(msg)
```
p-j-smith, line 156: I think if this exception is raised by the first call to `query_and_raise_if_exported`, the exception would be ignored and the query made again using MRN and accession number. Perhaps this could raise a different exception type to avoid querying the same exported image?

p-j-smith (approved):
looks good! And nice refactor of the tests. Very minor suggestion, feel free to ignore

pixl_dcmd/src/pixl_dcmd/_database.py
stefpiatek: Ah yeah that's not a bad point. I guess if we have a study uid and if its been exported then you're right, we shouldn't try again

PR #586, opened 2024-12-27
Strip whitespace and remove empty identifiers

stefpiatek:


cli/src/pixl_cli/_database.py
```diff
@@ -101,7 +101,7 @@ def _filter_exported_messages(
 ) -> pd.DataFrame:
     merged = messages_df.merge(
         images_df,
-        on=["accession_number", "mrn", "study_date"],
+        on=["accession_number", "mrn", "study_uid"],
```
p-j-smith, line 104: what would happen here for exported images in the db that have no `study_uid` (or is that even possible)? Would messages for these exported images be sent to the queue?

p-j-smith (approved):
ooo nice to see the reader more robust against input format

cli/src/pixl_cli/_database.py
stefpiatek: definitely happens with older ones, but they should match by accession number and mrn. The real fun is that for pinpoint alone we've got ones which were populated before we processed the study uid, but I've merged these in the database so that they have the uids already

stefpiatek: though ironically, I was doing some tidying and had to make good use of a backup in the processed 🤦🏻

PR #587, opened 2025-01-02
Log study info in anonymisation

stefpiatek:
Makes it easier to track processing of anonymisation


orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -310,36 +312,16 @@ def get_study_zip_archive_from_raw(resource_id: str) -> BytesIO:
     return BytesIO(response.content)
 
 
-def _get_study_resource_id(study_uid: str) -> str:
```
stefpiatek, line 313: FYI just moved down

orthanc/orthanc-anon/plugin/pixl.py
```diff
@@ -246,7 +248,8 @@ def _import_studies_from_raw(
     anonymised_study_uids = []
 
     for study_resource_id, study_uid in zip(study_resource_ids, study_uids, strict=False):
-        anonymised_uid = _anonymise_study_and_upload(study_resource_id, study_uid, project_name)
+        logger.debug("Processing project '{}', study '{}' ", project_name, study_uid)
```
p-j-smith, line 251: ah yeah, makes sense to log something here

p-j-smith (approved):
nice! Good idea to get the `StudyInfo` for logging / exceptions

PR #592, opened 2025-01-10
Changes after reviewing the PET data for DESPIAD

davecash75:
## Description
This builds the tag operations config file for the DESPIAD project and makes a few suggested changes for PET and CT. 

## Type of change
Please delete options accordingly to the description. These include:
* **Including more CT acquisition parameters in the main ct.yaml** - these have minimal, if any chance of containing personal information, but even though they are level 3, they might have some use for researchers.
* **Including more PET acquisition parameters in the main pet.yaml** - Height and Weight are routinely used in PET to compute Standardized Uptake Value. More information about the acquisition (reconstrution, randoms correction, etc) is also useful to ensure study protocol compliance in single and multi-centre studies. Acquisition time is a Level 1 attribute for PET series. And I think it may be worth keeping time to ensure uptake time, especially as in the PET base module all of the dates are either replace or wiped out. I also removed the DateTime attribute that was currently included in the Radiopharmaceutical Administration, as if you are keeping all dates out or replacing them, you wouldn't want this to leak out. If groups are doing PET scans that span midnight, then maybe it could be put back in as a separate yaml override?
* **Project specific DESPIAD tags** - My current understanding from our discussions is that given the study protocol and that we have informed in the ethics and in the consent that we will be collecting date of birth as well as dates of scan are part of the eCRF and the information we already hold, then this would be fine. Station Name is also helpful to check if all scans for a single subject are done on the same PET-CT camera, which is important for longitudinal analysis. 

- [X] New feature (non-breaking change which adds functionality)

### Suggested Checklist

### DIFF
--- a/projects/configs/tag-operations/pet.yaml
+++ b/projects/configs/tag-operations/pet.yaml
@@ -23,11 +23,11 @@
 - name: Series Time
   group: 0x0008
   element: 0x0031
-  op: replace
+  op: keep
 - name: Acquisition Time
   group: 0x0008
   element: 0x0032
-  op: replace
+  op: keep
 - name: Code Value
   group: 0x0008
   element: 0x0100
@@ -52,6 +52,14 @@
   group: 0x0008
   element: 0x010f
   op: keep
+- name: Patients Size
+  group: 0x0010
+  element: 0x1020
+  op: keep	
+- name: Patients Weight
+  group: 0x0010
+  element: 0x1030
+  op: keep
 - name: Radiopharmaceutical
   group: 0x0018
   element: 0x0031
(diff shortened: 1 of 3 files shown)


davecash75:
HI @p-j-smith I can't seem to assign reviewers, but would you please take a look?
THanks
Dave

projects/configs/tag-operations/pet.yaml
```diff
@@ -76,10 +88,6 @@
   group: 0x0018
   element: 0x1076
   op: keep
-- name: Radiopharmaceutical Start DateTime
-  group: 0x0018
-  element: 0x1078
-  op: keep
```
p-j-smith, line 82: ah yep, we shouldn't be keeping this value. The attribute is [required by the spec](https://dicom.innolitics.com/ciods/enhanced-pet-image/enhanced-pet-isotope/00540016/00181078) but we can `replace` rather than `keep` the value

p-j-smith (approved):
thanks for the explanations @davecash75 - happy to merge into my branch and have it all reviewed there before merging to main

PR #593, opened 2025-01-14
Enables PIXL DB to operate as a service separate from Orthanc Raw

### LINKED ISSUE #517: Allow PIXL DB to be run on a different database

As a system administrator, I'd like to make sure that PIXL DB (`pipeline` schema where we store `Extracts` and `Images`) can be configured to run on another database, rather than one running in docker so that it can be backed up.

### Testing

- [ ] Extra env variables for testing should just use the docker database

### Documentation

*No response*

### Dependencies

*No response*

### Details and Comments

Will need to alter the configuration to ensure that there is a `PIXL_DB` set of environmental variables, and then another set for `ORTHANC_RAW_DB`. In testing and development these can be set to the same docker postgres instance.

tomaroberts:
## Description
Fixes #517: 

Adds the means to create two postgres services. Previously, PIXL DB was used by both Orthanc Raw and for storing patient metadata. 

Now it is possible to have separate databases for Orthanc Raw and PIXL DB. By default and in keeping with previous design, the root `./docker-compose.yml` brings up a single postgres service, shared by both Orthanc Raw and PIXL DB. 

With this PR, `test/docker-compose.yml` now contains a new postgres service demonstrating how to bring up a separate service for PIXL DB.

## Type of change

- [x] This change requires a documentation update

### Suggested Checklist

- [x] I have passed on my local host device. (see further details at the [CONTRIBUTING](https://github.com/SAFEHR-data/PIXL/blob/main/CONTRIBUTING.md#local-testing) document)
- [x] Make sure your branch is up-to-date with main branch. See [CONTRIBUTING](https://github.com/SAFEHR-data/PIXL/blob/main/CONTRIBUTING.md) for a general example to syncronise your branch with the `main` branch.
- [x] I have requested review to this PR.
- [x] I have addressed and marked as resolved all the review comments in my PR.
- [x] Finally, I have selected `squash and merge`


p-j-smith (approved):
looking very nice!

I think the docs could do with updating in a couple of places:

- https://github.com/SAFEHR-data/PIXL/tree/main/cli#configuration
- https://github.com/SAFEHR-data/PIXL/tree/main/orthanc/orthanc-raw#step-1

And possibly some more places too. Which makes me wonder whether it would be worth describing all the environment variables in a single README rather than across the repo (as a separate PR)?

Also, if we use an external postgres instance, would we need to change how we handle [alembic migrations](https://github.com/SAFEHR-data/PIXL/tree/main/pixl_imaging/alembic#running-migrations-in-production)? Currently we run them when the container is started, but I guess we'd need to apply them manually if postgres is running elsewhere?

.env.sample
```diff
@@ -10,9 +10,17 @@ PIXL_DB_USER=pixl
...
+ORTHANC_RAW_DB_HOST=postgres
+ORTHANC_RAW_DB_PORT=5432
+ORTHANC_RAW_DB_NAME=pixl
+ORTHANC_RAW_DB_USER=pixl
+ORTHANC_RAW_DB_PASSWORD=
+
 # Exposed ports
 HASHER_API_PORT=
-POSTGRES_PORT=
+POSTGRES_PIXL_DB_PORT=
```
stefpiatek, line 22: Probably worth moving this elsewhere in the file as this port isn't necessarily what we're exposing it to (we can override this in the system test files, where we are exposing it).

cli/README.md
```diff
@@ -44,8 +44,8 @@ RABBITMQ_PORT=7008
 RABBITMQ_USERNAME=rabbitmq_username
 RABBITMQ_PASSWORD=rabbitmq_password
 
-POSTGRES_HOST=localhost
-POSTGRES_PORT=7001
+POSTGRES_PIXL_DB_HOST=localhost
+POSTGRES_PIXL_DB_PORT=7001
```
stefpiatek, line 48: How about `EXTERNAL_PIXL_DB_HOST` or `CLI_PIXL_DB_HOST` etc?

test/.env
```diff
@@ -29,7 +37,7 @@ RABBITMQ_ADMIN_PORT=7009
 FTP_PORT=20021
 
 # PIXL Export API
-POSTGRES_HOST=localhost
+POSTGRES_PIXL_DB_HOST=localhost
```
stefpiatek, line 40: I know it wasn't you but should this move? Feels like its not to do with the export api

stefpiatek (approved):
Looking good, thanks for getting this in 🎉 . Agree with Paul on the documentation places. Happy for you to merge once you've added that in (without another review) and resolved the other comments to your liking.

> Currently we run them when the container is started, but I guess we'd need to apply them manually if postgres is running elsewhere?

I think its fine for us to still run migrations on imaging api going up, because it'll still need to check the migrations have happened anyway.

test/docker-compose.yml
```diff
@@ -73,3 +74,29 @@ services:
...
+        build:
+            context: ../
+            dockerfile: ./docker/postgres/Dockerfile
+        environment:
+            POSTGRES_USER: ${PIXL_DB_USER}
+            POSTGRES_PASSWORD: ${PIXL_DB_PASSWORD}
+            POSTGRES_DB: ${PIXL_DB_NAME}
+            PGTZ: ${TZ:-Europe/London}
+        env_file:
+            - ../docker/common.env
```
stefpiatek, line 87: Imagine you've seen but from CI: `/home/runner/work/PIXL/docker/common.env: no such file or directory`

tomaroberts:
A note: my last commit removed the `env_file:` line from `test/docker-compose.yml` and now the system test is passing. But that's slightly unexpected?

Two thoughts:
1) I think the original problem was due to relative file paths. [See here for a similar issue](https://github.com/docker/compose/issues/11369) around relative paths and `env-file:`. I think the `context:` attribute is only applied to `build:` rather than the location of the file associated with `env-file:` attribute.
2) [Reading the DC docs](https://docs.docker.com/compose/how-tos/environment-variables/set-environment-variables/#use-the-env_file-attribute), there is some level of inheritance of env variables across compose files – this might explain the tests passing now because the env variables already exist and by removing the lines in [502f27e](https://github.com/SAFEHR-data/PIXL/pull/593/commits/502f27e372f06c2e0b3f000e076affad64595568) we remedy the broken file pathing.

stefpiatek:
> 1. think the original problem was due to relative file paths. [See here for a similar issue](https://github.com/docker/compose/issues/11369) around relative paths and `env-file:`. I think the `context:` attribute is only applied to `build:` rather than the location of the file associated with `env-file:` attribute.

Yeah I think that's right - IIRC we have to do the test `docker compose up --build` in two steps because of this

tomaroberts:
@stefpiatek – are you happy if we just remove the offending lines from the `test/docker-compose.yml` then? As per https://github.com/SAFEHR-data/PIXL/pull/593/commits/502f27e372f06c2e0b3f000e076affad64595568

.env.sample
tomaroberts: Done L13-14: https://github.com/SAFEHR-data/PIXL/blob/b8d5ec35ce2299018f3f83382be51f8f9379d982/.env.sample#L13

test/.env
tomaroberts: Done L13-14: https://github.com/SAFEHR-data/PIXL/blob/b8d5ec35ce2299018f3f83382be51f8f9379d982/.env.sample#L13-L14

cli/README.md
tomaroberts: Sticking with `POSTGRES_ORTHANC_RAW_DB_PORT` and `POSTGRES_PIXL_DB_HOST`. Will explain in comment.

tomaroberts:
@stefpiatek 

@p-j-smith and I have discussed at (surprising) length how best to name the environment variables as it's fiddly and quite confusing. This simultaneously feels like making a mountain out of a molehill, so please excuse the lengthy comment, but also we're mindful of causing confusion further down the line for someone who might want to setup an external PIXL DB.

We think that having the variable names `POSTGRES_ORTHANC_RAW_DB_PORT` and `POSTGRES_PIXL_DB_PORT` is optimal (albeit imperfect). With these variables, we are trying to differentiate between the postgres dbs associated with Orthanc Raw and PIXL DB. 

We figured that your suggestions of using:
- `EXTERNAL_PIXL_DB_POST` sounds odd where it is called inside [_config.py](https://github.com/SAFEHR-data/PIXL/blob/f50d4312160c97e45431468d6a72d36d4f79ca51/cli/src/pixl_cli/_config.py#L32) because it might not be external
- Similarly calling it `CLI_PIXL_DB_PORT` doesn't feel explicit enough that the variable is about postgres and also for consistency, this would mean renaming `POSTGRES_ORTHANC_RAW_DB_PORT` > `ORTHANC_RAW_DB_PORT`, which is already something else 😱 

**So...**

I've added a comment to the `.env.sample` to hopefully make it clear that the ports should both be 7001 for a purely local deployment and then `POSTGRES_PIXL_DB_PORT` can become 7011 if it is required as an external db.

tomaroberts:
@p-j-smith – I changed the README documentation you mentioned. I noticed that the READMEs are quite selective in which env variables they specify compared to the `.sample.env`, so just re-requested your review to check I've included enough.

Otherwise... I think we're good to go?!

.env.sample
```diff
@@ -5,14 +5,26 @@ CLI_RETRY_SECONDS=300
...
+ORTHANC_RAW_DB_PORT=5432
+ORTHANC_RAW_DB_NAME=pixl
+ORTHANC_RAW_DB_USER=pixl
+ORTHANC_RAW_DB_PASSWORD=
+
 # Exposed ports
 HASHER_API_PORT=
-POSTGRES_PORT=
+POSTGRES_ORTHANC_RAW_DB_PORT=
+POSTGRES_PIXL_DB_PORT= # equals POSTGRES_ORTHANC_RAW_DB_PORT unless require external PIXL_DB service
```
stefpiatek, line 27: I'm a bit confused though, the only places that this is ever used is in the external port mapping, and its not to do with orthanc raw at all. Or am I misunderstanding this? This feels like a description of what its doing 

```suggestion
POSTGRES_EXTERNALLY_MAPPED_DB_PORT=
POSTGRES_PIXL_DB_PORT= # equals POSTGRES_EXTERNALLY_MAPPED_DB_PORT unless require external PIXL_DB service
```

image

p-j-smith: yep I see how this is super confusing! I think ideally we could use a single variable like `EXTERNAL_PIXL_DB_PORT `, but the issue comes in when running the system tests and we have two postgres containers running. On both containers we're opening a port to postgres:
https://github.com/SAFEHR-data/PIXL/blob/325d938704b6039721e9ea6a766dbf04150695cd/docker-compose.yml#L381
https://github.com/SAFEHR-data/PIXL/blob/325d938704b6039721e9ea6a766dbf04150695cd/test/docker-compose.yml#L92
And it can't be the same port we expose as the containers are running on the same host. We don't actually need to expose the `POSTGRES_ORTHANC_RAW_DB_PORT ` if the PIXL db is running externally, but I'm not sure if there's a way to conditionally open a port? If not, then we were thinking we need some way in the system tests to distinguish what the two exposed ports are for, and the only thing we could think of was what postgres is used for in each container. I'm sure there are better names though (or perhaps a better way of doing it?)

tomaroberts: Sorry, it is confusing, I agree. My understanding below, unless @p-j-smith correct me if I'm wrong 😆 

**POSTGRES_ORTHANC_RAW_DB_PORT**

[Line 381](https://github.com/SAFEHR-data/PIXL/blob/325d938704b6039721e9ea6a766dbf04150695cd/docker-compose.yml#L381) in the **root** docker-compose.yml is within the `postgres` service. This service is used as a db for Orthanc Raw on port 7001.

**POSTGRES_PIXL_DB_PORT**

Next, `POSTGRES_PIXL_DB_PORT` is used in pixl_cli `_config.py` [L33](https://github.com/SAFEHR-data/PIXL/blob/325d938704b6039721e9ea6a766dbf04150695cd/cli/src/pixl_cli/_config.py#L33) where if it is 7001 then both Orthanc Raw and PIXL DB use the same `postgres` service.

If we set it to an arbitrary port, e.g. 7011, this is used on [L92](https://github.com/SAFEHR-data/PIXL/blob/325d938704b6039721e9ea6a766dbf04150695cd/test/docker-compose.yml#L92) in the `postgres-pixl-db` service within `test/docker-compose.yml` which demonstrates how to setup PIXL_DB as a separated from Orthanc Raw (which remains on 7001).

stefpiatek: I guess I'm not advocating for having a single env variable, just renaming it so that it makes sense in a production environment. If we end up having something strange in the system test setup where it doesn't quite make sense and requires extra comments so say what's going on then I'm fine with that. Does my suggestion work in that case?

stefpiatek: >  We don't actually need to expose the `POSTGRES_ORTHANC_RAW_DB_PORT ` if the PIXL db is running externally, but I'm not sure if there's a way to conditionally open a port? 

We could use [docker compose profiles](https://docs.docker.com/compose/how-tos/profiles/) to do this actually. so in the main docker compose we have postgres-exposed which has profiles for `pixl_dev` and `pixl_test`. then we have a service for postgres for profiles `system-test` and `pixl_prod`?

tomaroberts: > I guess I'm not advocating for having a single env variable, just renaming it so that it makes sense in a production environment.

Yeah, totally agree and get where you're coming from.

Paul also suggested docker compose profiles yesterday too. Let me have a look into it.

tomaroberts:
I'm going to _tentatively_ say this is done now 🥶, if you guys wouldn't mind taking a look!

cli/README.md
```diff
@@ -20,20 +20,42 @@ uv sync
...
+**2) Start-up with External PIXL DB**
+Set the port number environment variable of your external PIXL DB:
 ```bash
-pixl --help
+CLI_PIXL_DB_PORT=7001
+```
+
+Start-up PIXL:
+```bash
+pixl dc up --profile postgres
```
stefpiatek, line 53: Can I suggest instead that we have an environmental variable that the CLI reads and works out if it should be exposed or not? I don't want to have to always remember when re-upping production whether it should be exposed or not.

Maybe?
```
EXTERNAL_PIXL_DB=TRUE
```

stefpiatek (approved):
Looks really nice, one request to have the profile work from config and not from having to remember to call the right profile. Once you've done that I'm happy for you to merge 😄

cli/README.md
p-j-smith: oh that's a much better idea

cli/src/pixl_cli/_docker_commands.py
```diff
@@ -42,11 +43,17 @@ def dc(args: tuple[str]) -> None:
...
 def _parse_up_args(args: tuple[str, ...]) -> list:
-    """Check up args and define default profile if not set"""
+    """Check up args and set docker compose profile"""
     args_list = list(args)
-    if "--profile" not in args:
-        up_index = args.index("up")
-        args_list[up_index:up_index] = ["--profile", "postgres-exposed"]
+
+    up_index = args.index("up")
+    external_pixl_db_env = os.environ.get("EXTERNAL_PIXL_DB", "false").lower()
```
stefpiatek, line 50: I think we're using decouple's config elsewhere for this so probably worth using that for consistency 😄

tomaroberts: Ah, orthanc-anon and orthanc-raw plugins use os.environ.get() but I'll amend to decouple, which will be consistent within pixl_cli 👍

tomaroberts:
I will merge tomorrow, unless someone spots something in the meantime.

Thanks for your help both, especially Paul – owe you a 🍺 / 🍸 / ☕ (delete as preferred).

p-j-smith (approved):
nice work @tomaroberts !

> Thanks for your help both, especially Paul – owe you a 🍺 / 🍸 / ☕ (delete as preferred).

Ooo I do like me a nice ☕

cli/src/pixl_cli/_docker_commands.py
stefpiatek: Oh interesting that they do, we should probably also change that but that doesn't need to happen this PR

PR #595, opened 2025-01-28
Docker Compose profile for external PIXL DB

tomaroberts:
@p-j-smith, @stefpiatek – draft PR alternative approach to #593, to hopefully demonstrate this is along the right lines.

Uses Docker Compose profiles approach, as discussed in that PR. Thought simpler to start again from `main` rather than unpick the other one.

I've got two postgres containers running with `./run-system-test.sh`, however the system-tests are failing hard. I can't figure out why yet. I think pixl cli is unable to communicate with system-test-postgres-external-1.

If either of you has a chance to pull the branch and investigate the system tests errors, let me know if you can diagnose, otherwise I'll keep at it.

### DIFF
--- a/cli/src/pixl_cli/_config.py
+++ b/cli/src/pixl_cli/_config.py
@@ -37,6 +37,9 @@ SERVICE_SETTINGS = {
     },
 }  # type: dict
 
+if config("POSTGRES_EXTERNAL_PORT"):
+    SERVICE_SETTINGS["postgres"]["port"] = int(config("POSTGRES_EXTERNAL_PORT"))
+
 
 class APIConfig:
     """API Configuration"""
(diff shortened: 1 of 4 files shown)


stefpiatek:
> If either of you has a chance to pull the branch and investigate the system tests errors, let me know if you can diagnose, otherwise I'll keep at it.

Looks like its passing in CI which is interesting.

tomaroberts:
> > If either of you has a chance to pull the branch and investigate the system tests errors, let me know if you can diagnose, otherwise I'll keep at it.
> 
> Looks like its passing in CI which is interesting.

The system-test was being skipped, which is what was failing.

Paul and I reverted to my previous PR/branch, so don't worry about this draft now.

PR #598, opened 2025-02-13
Add support for Docker Compose<22.0

### LINKED ISSUE #597: `pixl dc up` fails with Docker Compose versions <2.20 as the `required` flag is not supported
### What happened?

In #593, `required: false` was set for some of the `depends_on` attributes in `docker-compose.yml`. This was necessary as 
 
However, `required` is [only supported in Docker Compose>=2.20.0](https://docs.docker.com/reference/compose-file/services/#long-syntax-1). To support older versions of Docker Compose, we could:

- remove use of the `required` flag
- remove use of profiles - remove the `postgres` service and rename `postgres-exposed` to `postgres` (i.e. the same situation we had before #593, always exposing `CLI_PIXL_DB_PORT`)
- don't spin up an second postgres container for the integration tests

The second point is necessary as `required` defaults to `true`, meaning if we only remove the `required` flag the services that depend on `postgres`/`postgres-exposed` would fail to start when using Docker Compose>=2.20.

### Relevant log output (optional)

_No response_

### What did you expect? (optional)

_No response_

p-j-smith:
Fixes #597 

- Remove use of profiles with docker compose
- Always expose the `CLI_PIXL_DB_PORT` on the postgres container
- Remove `EXTERNAL_PIXL_DB` env var as we're no longer using profiles
- Remove second postgres container used in integration tests; use the same postgres instance for PIXL DB and Orthanc Raw
- We can still have PIXL DB running in an external postgres instance (by setting the relevant env vars), but we're not testing for this in CI
- Update docs to reflect changes
- tested on dev instance on GAE

## Description
Fixes #{issue_number}: A few words or sentences describing the changes proposed in this pull request (PR).

## Type of change

### Suggested Checklist

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -4,19 +4,18 @@ PIXL_QUERY_TIMEOUT=10
 CLI_RETRY_SECONDS=300
 
 # PIXL PostgreSQL instance
-PIXL_DB_HOST=postgres-exposed
+PIXL_DB_HOST=postgres
 PIXL_DB_PORT=5432
 PIXL_DB_NAME=pixl
 PIXL_DB_USER=pixl
 PIXL_DB_PASSWORD=
 SKIP_ALEMBIC=false
-EXTERNAL_PIXL_DB=false
 
 # PIXL DB Postgres host
 CLI_PIXL_DB_HOST=localhost
 
 # Orthanc Raw PostgreSQL instance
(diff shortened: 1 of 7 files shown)


tomaroberts (approved):
Looks good Paul and just like it used to! 😆

Go for it.

PR #600, opened 2025-05-29
Change pixl schema so that it makes sense on multi-schema db

stefpiatek:
We can't have it called `pixl` because for CI postgres doesn't like it when the database is the same name as the schema. So we had it as `pipeline` but that doesn't make sense on a shared database like the UDS


milanmlft (approved):
LGTM 👌

PR #624, opened 2026-02-20
Update minor versions except for dicom-validator

stefpiatek:
dicom-validator update will require changing how we pre-download the schema


p-j-smith (approved):
lovely stuff

PR #633, opened 2026-06-03
Use new layout of omop datasets

stefpiatek:
Update to new batched omop output, which has a `omop` subdirectory in the `public` folder


p-j-smith (approved):
looks good! Does this mean the any old extracts we have can no longer be exported with PIXL (without being restructured)?

stefpiatek:
Yeah that's right, we'd just need to add an omop sub directory

PR #634, opened 2026-06-03
Store TRE API zip archive in temp directory

stefpiatek:
So that it cleans up the archive after sending, currently these accumulate in the `pixl_core` directory.


p-j-smith (approved):
yeah it makes sense to clean these up, nice!

